PR #7, opened 2020-04-05
Removing packaging files from the repository

lazyfrosch:
Please refer to the VCS from any distribution to get their current
packaging files.

When we start doing packaging builds for users, we should create extra
repositories for the packaging code.

We did the same with @Icinga a long time ago, the packaging is basically dead code we include in our repository.


lazyfrosch:
See [Distributions in the Wiki](https://github.com/gnome-terminator/terminator/wiki/Distributions) for a list of downstreams

epozuelo:
Agreed.

PR #9, opened 2020-04-05
Converting to python 3 and making all tests pass.

JAugusto42:
It's not a big deal, but as it is part, maybe in the future we can improve these tests.

### DIFF
--- a/tests/test_doctests.py
+++ b/tests/test_doctests.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!/usr/bin/env python
 """Load up the tests."""
 
 import os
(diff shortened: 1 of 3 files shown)


tests/test_doctests.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
```
lazyfrosch, line 1: Please avoid modifying the shebang in this way. I discussed this with @mattrose in chat briefly.

Current best practice for Python (e.g. when using virtualenvs) is keeping the term `python` while selecting from `PATH`

In terms for packaging we need to work around at the moment, usually this is handled by standard packaging tools and setuptools.

Until I can improve the integration with setuptools (with 2.0) we could `sed` the shebang in packaging scripts.

JAugusto42: yes, i revert here.
thx

PR #15, opened 2020-04-12
update the INSTALL file

mattrose:
I just updated the INSTALL file to get rid of the old website, and update all of the instructions so that they, you know, work.


geckon:
BTW, there is also a package in openSUSE:
https://software.opensuse.org/package/terminator

I'm not the maintainer, just user though.

mattrose:
I updated the INSTALL file to reflect that, thanks.

PR #27, opened 2020-04-19
Remove obsolete terminator.wrapper

lazyfrosch:
We don't really need this file, if one needs to wrap the binary they could use `$HOME/bin/terminator`

Anyone thoughts on this?


DarthGandalf:
It's also in MANIFEST.in

lazyfrosch:
Thanks @DarthGandalf

Issue #10, opened 2020-04-07
VTE Regexp should work with older VTE releases as well


jest:
Since the commit b2a47a50fdcd867335960d35966dc4cebe4c33c0 the code is broken on Ubuntu 16.04, which uses the old VTE:

```
$ terminator
Traceback (most recent call last):
  File "/usr/bin/terminator", line 122, in <module>
    TERMINATOR.create_layout(OPTIONS.layout)
  File "/usr/share/terminator/terminatorlib/terminator.py", line 311, in create_layout
    window, terminal = self.new_window()
  File "/usr/share/terminator/terminatorlib/terminator.py", line 234, in new_window
    terminal = maker.make('Terminal')
  File "/usr/share/terminator/terminatorlib/factory.py", line 94, in make
    output = func(**kwargs)
  File "/usr/share/terminator/terminatorlib/factory.py", line 106, in make_terminal
    return(terminal.Terminal())
  File "/usr/share/terminator/terminatorlib/terminal.py", line 147, in __init__
    self.update_url_matches()
  File "/usr/share/terminator/terminatorlib/terminal.py", line 273, in update_url_matches
    reg = Vte.Regex.new_for_match(re, len(re), self.regex_flags)
  File "/usr/lib/python2.7/dist-packages/gi/module.py", line 139, in __getattr__
    self.__name__, name))
AttributeError: 'gi.repository.Vte' object has no attribute 'Regex'
```

This is reproducible by using PPA `ppa:gnome-terminator/nightly-gtk3` on 16.04, caused there by commit version `1.91-0~1835~ubuntu16.04.1`.

In https://bugs.launchpad.net/terminator/+bug/1867746/comments/7 you mentioned that the "real" development is moved to GitHub, so I feel encouraged to raise an issue here. :)

(also, I have no idea on how to sanely report the issue on the launchpad...)

Is 16.04 supposed to be dropped? If so, the aformentioned build on LP should be stopped on 1834 for Xenial.

lazyfrosch:
Thanks for your issue

This should have been fixed with #6 and is merged to master, we are working on a release soon.

We don't own the Launchpad project currently, as well as the PPA, so there are no "nightly" builds!

lazyfrosch:
Oh sorry I misread your issue, I will have a look at that!

jest:
@lazyfrosch the released version works on 16.04. However, on 18.04 I have problems:
```

[...]

PR #28, opened 2020-04-19
terminal: Improve compat for Vte Regex

lazyfrosch:
Try the latest Vte functions, and fall back to Glib when this doesn't work.

This should fix problems with Ubuntu 17.10 and 18.04, where they avoided PCRE2 for Vte.

Can you give it a try @jest?
 
Tested:
- [x] Ubuntu 16.04
- [x] Ubuntu 18.04
- [x] Debian 10 buster
- [x] Debian bullseye

followup for #10


jest:
Works great on 18.04, with multiline links too :+1:

Thanks!

PR #30, opened 2020-04-20
Update pt_BR.po

chclxds:
Updated Translation


lazyfrosch:
Thank you, we are going to handle translation with transifex.com in the future

chclxds:
> Thank you, we are going to handle translation with transifex.com in the future

Great news, I would like to contribute there, as I do with several projects, including XFCE and TOR

PR #35, opened 2020-04-21
fix exception when feeding terminal number to terminal

mattrose:
It's a weird feature that I've never heard of but in the titlebar menu in the top left corner there's two menu items "Insert terminal number" and "Insert padded terminal number".  These were not previously working and would just put the following to stdout.  This PR fixes this issue.

```
  File "/home/mattrose/Code/terminator/terminatorlib/terminator.py", line 642, in do_enumerate
    term.feed(numstr % (idx + 1))
  File "/home/mattrose/Code/terminator/terminatorlib/terminal.py", line 1541, in feed
    self.vte.feed_child(text, len(text))
TypeError: Vte.Terminal.feed_child() takes exactly 2 arguments (3 given)
```

### DIFF
--- a/terminatorlib/terminal.py
+++ b/terminatorlib/terminal.py
@@ -1538,7 +1538,7 @@ class Terminal(Gtk.VBox):
 
     def feed(self, text):
         """Feed the supplied text to VTE"""
-        self.vte.feed_child(text, len(text))
+        self.vte.feed_child_binary(text.encode(self.vte.get_encoding()))
 
     def zoom_in(self):
         """Increase the font size"""


mattrose:
NB: we'll likely want to change this back once https://gitlab.gnome.org/GNOME/vte/-/issues/201 gets fixed

lazyfrosch:
Hm feeding binary it not too bad, will review it later.

mattrose:
It's the same way I fixed the other instance of feed_child.  the feed_child_binary thing is a workaround for  the vte issue.

PR #36, opened 2020-04-22
fast resize keyboard shortcuts

waldner:
This is the patch originally proposed and discussed on LP here: https://bugs.launchpad.net/terminator/+bug/1695935 (code here: https://code.launchpad.net/~waldner/terminator/fast-resize) which was stuck for lack of feedback from previous maintainers.

As explained there, there is the option to keep the new "fast resize" bindings entirely separate from the normal ones, or use only the existing keybindings but make the resize step configurable (leaving it as 1 by default as to not break existing setups). The patch implements the former strategy, but I'm happy to modify it to implement the latter.


mattrose:
Given that other people have applied this PR to their own tree, I'm gonna merge this in.  I may not think it's all that useful, but I'm willing to accept I may be wrong in this case.

Issue #38, opened 2020-04-25
[Feature Request] Make bold-is-bright option configurable


JakubVanek:
Hi all!

I've recently switched to Ubuntu 20.04 and I was surprised that my terminal prompt looks darker. This is a problem for me - I have problems seeing small letters with reduced contrast.

It turns out that GNOME Terminal (and VTE) have switched the bold-is-bright option to off by default: https://askubuntu.com/a/1156969

One of the ways forward (for me) is to change PS1 and LS_COLORS. On the other hand, just switching the option back to its previous value is easier. The issue is that Terminator does not provide the checkbox that gnome-terminal has.

Terminator currently does not touch this setting, VTE therefore uses the default. However, by adding `self.vte.set_bold_is_bright(True)` after the following line, I can revert VTE to the old behaviour.

https://github.com/gnome-terminator/terminator/blob/11a6a57f517df443e181c87970da2ef7d990b3a3/terminatorlib/terminal.py#L703

This would then have to be plumbed through the configuration UI.

Would you please consider this?

Best regards,

Jakub

BachoSeven:
This is useful also when setting custom dircolors themes

DarthGandalf:
Thanks for this! I was wondering why sometimes it shows black text on black background.

PartialVolume:
Thanks for this fix, we just discovered this was a problem with [nwipe](https://github.com/martijnvanbrummelen/nwipe/issues/359) in a minority of terminals

Glad this option is now available, thanks.

PR #39, opened 2020-04-25
#38: Add option for configuring bold-is-bright setting of VTE

JakubVanek:
This PR implements the suggested bold-is-bright checkbox.

Appearance:
Settings UI

Fixes https://github.com/gnome-terminator/terminator/issues/38


JakubVanek:
There's one element of this that should be discussed - do we want to follow gnome-terminal and disable bold-is-bright by default? It is a user-facing change, but it is keeping in line with others.

mattrose:
This looks really good.  Thanks for all your hard work

PR #41, opened 2020-04-25
#40: remove default key bindings for input broadcasting

### LINKED ISSUE #40: Remove default keybindings for enabling terminal broadcast
Activating terminal broadcast through a key combination could prove very dangerous in certain situations, see previous issue: https://bugs.launchpad.net/terminator/+bug/1872460

JakubVanek:
Fixes #40. This PR removes the predefined `Alt+A`, `Alt+G` and `Alt+O` broadcast key bindings from default configuration. Therefore users will have to set the shortcuts manually.

However, I'm not entirely sure this is the best way to handle this, perhaps a warning window could pop up the first time such a key combination is pressed.


FernandoBasso:
Why do you want to remove those keybindings? What is the purpose and what are the benefits of doing so?

FernandoBasso:
I think good arguments were made in favor or removing those keybindings by default. I agree that we should disable them (by default).

ohir:
@lazyfrosch 
> *Markus Frosch (lazyfrosch) 12 hours ago*
> Changed in terminator:
> status:	New → Won't Fix

May you elaborate more on (launchpad's) status 'Wontfix'?

lazyfrosch:
@ohir there is no status "moved to somewhere" else, and bugs are no longer handled on LP

PR #42, opened 2020-04-26
Add layout launcher to terminal popup menu.

nbeaver:
I often want to launch a layout, but can't remember what the keyboard shortcut is. This patch adds it to the right-click popup menu.

Older patch on Launchpad:

https://bugs.launchpad.net/terminator/+bug/1800735

### DIFF
--- a/terminatorlib/terminal_popup_menu.py
+++ b/terminatorlib/terminal_popup_menu.py
@@ -13,6 +13,7 @@ from .util import err, dbg
 from .config import Config
 from .prefseditor import PrefsEditor
 from . import plugin
+from layoutlauncher import LayoutLauncher
 
 class TerminalPopupMenu(object):
     """Class implementing the Terminal context menu"""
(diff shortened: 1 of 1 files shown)


mattrose:
Thanks for this patch, and thank you for reviewing it @JakubVanek.  I'm gonna pull in this request if it works for me, and I thought of an even better idea that might replace it.

mattrose:
So I tried to fix this up, but failed.  It just needs to import the layoutlauncher file properly so that terminator doesn't crash on Python3.  Can you fix this up?

terminatorlib/terminal_popup_menu.py
```diff
@@ -13,6 +13,7 @@
 from .config import Config
 from .prefseditor import PrefsEditor
 from . import plugin
+from layoutlauncher import LayoutLauncher
```
mattrose, line 16: this neeeds to be from `from .layoutlauncher import LayoutLauncher`

nbeaver: Sounds good to me. I don't really know much about how Github works; do I need to do anything?

lazyfrosch:
```suggestion
from .layoutlauncher import LayoutLauncher
```

PR #46, opened 2020-04-29
Added option to disable ctrl+mousewheel zoom

filipkilibarda:
Addressing the longstanding <kbd>Ctrl+scroll</kbd> zoom in/out issue, as expressed in these old requests
- https://bugs.launchpad.net/terminator/+bug/1505292
- https://bugs.launchpad.net/terminator/+bug/1638607

## What problem does this PR solve?
Some laptop users (including myself) often bump the trackpad while holding <kbd>Ctrl</kbd>, which results in the terminal drastically zooming in and out (it's super sensitive).

The usual fix to get around this was to press <kbd>Ctrl+0</kbd> to reset the zoom everytime it happened, but this often fails to restore the original symmetry among split windows in apps like `vim` and `tmux`.

Needless to say, it's also annoying having to reset zoom scale so often.

## What this PR does
- Adds a checkbox to the `Profiles > General` section of terminator preferences
- If checked, don't run the <kbd>Ctrl+mousewheel</kbd> signal handler that's responsible for zooming the terminal in and out
- keyboard bindings for zooming still work
---
Thanks for taking over maintenance :) Cool to see this project active again!

### DIFF
--- a/terminatorlib/config.py
+++ b/terminatorlib/config.py
@@ -118,7 +118,6 @@ DEFAULTS = {
             'title_font'            : 'Sans 9',
             'putty_paste_style'     : False,
             'smart_copy'            : True,
-            'disable_mousewheel_zoom': False,
         },
         'keybindings': {
             'zoom_in'          : '<Control>plus',
@@ -222,6 +221,7 @@ DEFAULTS = {
                 'scroll_on_output'      : False,
                 'scrollback_lines'      : 500,
                 'scrollback_infinite'   : False,
+                'disable_mousewheel_zoom': False,
                 'exit_action'           : 'close',
                 'palette'               : '#2e3436:#cc0000:#4e9a06:#c4a000:\
 #3465a4:#75507b:#06989a:#d3d7cf:#555753:#ef2929:#8ae234:#fce94f:\
--- a/terminatorlib/config.py
+++ b/terminatorlib/config.py
@@ -118,6 +118,7 @@ DEFAULTS = {
             'title_font'            : 'Sans 9',
             'putty_paste_style'     : False,
             'smart_copy'            : True,
+            'disable_mousewheel_zoom': False,
         },
         'keybindings': {
             'zoom_in'          : '<Control>plus',
--- a/terminatorlib/terminal.py
+++ b/terminatorlib/terminal.py
@@ -972,12 +972,13 @@ class Terminal(Gtk.VBox):
 
     def on_mousewheel(self, widget, event):
         """Handler for modifier + mouse wheel scroll events"""
-        if self.config["disable_mousewheel_zoom"] is True:
-            return (True)
         SMOOTH_SCROLL_UP = event.direction == Gdk.ScrollDirection.SMOOTH and event.delta_y <= 0.
         SMOOTH_SCROLL_DOWN = event.direction == Gdk.ScrollDirection.SMOOTH and event.delta_y > 0.
         if event.state & Gdk.ModifierType.CONTROL_MASK == Gdk.ModifierType.CONTROL_MASK:
-            # Ctrl + mouse wheel up/down with Shift and Super additions
+            # Zoom the terminal(s) in or out if not disabled in config
+            if self.config["disable_mousewheel_zoom"] is True:
+                return (False)
+            # Choice of target terminals depends on Shift and Super modifiers
             if event.state & Gdk.ModifierType.MOD4_MASK == Gdk.ModifierType.MOD4_MASK:
                 targets=self.terminator.terminals
             elif event.state & Gdk.ModifierType.SHIFT_MASK == Gdk.ModifierType.SHIFT_MASK:
--- a/terminatorlib/terminal.py
+++ b/terminatorlib/terminal.py
@@ -972,6 +972,8 @@ class Terminal(Gtk.VBox):
 
     def on_mousewheel(self, widget, event):
         """Handler for modifier + mouse wheel scroll events"""
+        if self.config["disable_mousewheel_zoom"] is True:
+            return (True)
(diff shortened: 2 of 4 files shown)


terminatorlib/terminal.py
```diff
@@ -972,6 +972,8 @@ def on_buttonpress(self, widget, event):
 
     def on_mousewheel(self, widget, event):
         """Handler for modifier + mouse wheel scroll events"""
+        if self.config["disable_mousewheel_zoom"] is True:
+            return (True)
```
JakubVanek, line 976: This handler is common for all scrolling actions. This therefore also disables scrolling. This should be checked a few lines below where modifiers are checked.

terminatorlib/config.py
```diff
@@ -118,6 +118,7 @@
             'title_font'            : 'Sans 9',
             'putty_paste_style'     : False,
             'smart_copy'            : True,
+            'disable_mousewheel_zoom': False,
```
JakubVanek, line 121: I think this should be a profile-level setting, but I'm not sure if it fits more into the Scrolling or the General sections.

mckellyln:
@JakubVanek hi
I cannot seem to get Alt + scrollwheel to work in this terminator.  
I just wanted to ask would that be in this same area of the code ?
thx,
-m

JakubVanek:
Hi @mckellyln,

yup, the following patch prints ALT SCROLL for me:
```patch
diff --git a/terminatorlib/terminal.py b/terminatorlib/terminal.py
index 64a14bdd..0dd36570 100644
--- a/terminatorlib/terminal.py
+++ b/terminatorlib/terminal.py
@@ -1006,6 +1006,8 @@ class Terminal(Gtk.VBox):
             elif event.direction == Gdk.ScrollDirection.DOWN or SMOOTH_SCROLL_DOWN:
                 self.scroll_by_page(1)
                 return (True)
+        if event.state & Gdk.ModifierType.MOD1_MASK == Gdk.ModifierType.MOD1_MASK:
+            print("ALT SCROLL")
         return(False)
 
     def popup_menu(self, widget, event=None):
```

There is no ALT_MASK, but according to https://valadoc.org/gdk-3.0/Gdk.ModifierType.html the corresponding mask is mainly MOD1_MASK.

Best regards,
Jakub

mckellyln:
Thanks.  I do not get that.  Something in my setup is affecting alt + wheel in my terminator.
Other apps work fine so its not the OS.  I'll keep digging.
I really appreciate everyone's help on the great Terminator!

JakubVanek:
I cannot scroll in terminator either if I have Alt pressed, so the issue is not only on your machine. However I suspect that this may not be a change in Terminator, but rather in GNOME VTE - that is a widget that effectively does the terminal emulation. It could be worked around in terminator, but some nice-to-have features may be lost (like smooth scrolling on a touchpad) (this is only a pure speculation though).

mckellyln:
ok, thanks for that good info.
I will research if I can somehow re-configure my gnome/vte to allow mod1 + wheel ?
-m

terminatorlib/terminal.py
filipkilibarda: Ah yes, I definitely forgot to think here.

terminatorlib/config.py
filipkilibarda: Yeah agreed, I'll think about it

filipkilibarda: I think "general" might be more suitable.

Every other option in the "Scrolling" section is related to scrolling up and down through the content of a terminal, which is not what this PR is about.

JakubVanek:
Curiously, with zoom disabled, scrolling with Ctrl pressed seems to work fine. There seems to be something special about the Alt key.

filipkilibarda:
Here's an ugly patch that makes it so you can scroll while holding <kbd>Alt</kbd>. Not sure if this is what @mckellyln meant

```diff
diff --git a/terminatorlib/terminal.py b/terminatorlib/terminal.py
index e8212197..1a67f32c 100644
--- a/terminatorlib/terminal.py
+++ b/terminatorlib/terminal.py
@@ -1001,6 +1001,8 @@ class Terminal(Gtk.VBox):
             elif event.direction == Gdk.ScrollDirection.DOWN or SMOOTH_SCROLL_DOWN:
                 self.scroll_by_page(1)
                 return (True)
+        if event.state & Gdk.ModifierType.MOD1_MASK == Gdk.ModifierType.MOD1_MASK:
+            event.state ^= Gdk.ModifierType.MOD1_MASK
         return(False)

     def popup_menu(self, widget, event=None):
```

There's some signal handler in the chain (not in terminator - maybe libvte?) that aborts or ignores scrolling when <kbd>Alt</kbd> is pressed, so quick hack is to just turn off <kbd>Alt</kbd> in our signal handler, which comes first in the chain.

mckellyln:
Thanks!  I will test.  What I did before is scroll 4x more with Alt+wheel then just with wheel.
So in vi or tmux or whatever I can tell alt mod is pressed while wheel is moved.

filipkilibarda:
> Update: ok, I see that you removed the alt - so scroll works but we don't know alt was pressed.
> makes sense, it is better than not scrolling at all! 

Yeah I think we would want a better solution to make this configurable. Just killing the <kbd>Alt</kbd> modifier seems too invasive.

Any further research would require digging through libgtk or libvte to see why <kbd>Alt</kbd> is handled differently.

Glad it works better though

mattrose:
I'll take a look at this in the next few days.

mattrose:
Found this related thing while scrolling through VTE issues:

https://gitlab.gnome.org/GNOME/vte/-/issues/241

mattrose:
I left some feedback on the related vte bug mentioned above based on comments here and in the issue that @mckellyln raised.

PR #49, opened 2020-05-01
Make failed tests fail the test

DarthGandalf:


lazyfrosch:
This script always was weird for me, I will move testing over to pytest.

Thanks for noticing the problem!

lazyfrosch:
I will address your commit in #50, where I also fix the tests.

Thank you!

PR #50, opened 2020-05-01
Switch to setuptools and use pytest

### LINKED ISSUE #14: Switch to setuptools
To better handle installation and dependencies we should switch from `distutils` to `setuptools`.

This will help with packaging too.

lazyfrosch:
Also fix some minor issues with DocTests

fixes #14 
fixes #49

Thanks to @DarthGandalf


lazyfrosch:
Open issues on testing seem to be caused by not having a display in CI.

lazyfrosch:
I guess we should use this: https://pygobject.readthedocs.io/en/latest/guide/testing.html

> To get automated tests of GTK code running on a headless server use Xvfb (virtual framebuffer X server). It provides the xvfb-run -a command which creates a temporary X server without the need for any real display hardware.
>
> `xvfb-run -a python my_script.py`

lazyfrosch:
I disabled Python 3.6 and 3.7 for now, they seem to segfault due to unrelated reasons.

PR #51, opened 2020-05-01
Disable special logic for pasting on Wayland

### LINKED ISSUE #24: Middle click does not paste selected text
Setup: CentOS 8 / Gnome (Wayland)

Steps to reproduce:

1. Open a gnome terminal and highlight some text with the mouse.
2. Middle click within the gnome terminal and observe that the highlighted text is pasted.
3. Middle click within a terminator window and observe that the pasted contents are from a different clipboard (not the highlighted text).

lazyfrosch:
Patch by @fonzinc

fixes #24

### DIFF
--- a/terminatorlib/terminal.py
+++ b/terminatorlib/terminal.py
@@ -927,12 +927,11 @@ class Terminal(Gtk.VBox):
             # Suppress double-click behavior
             return True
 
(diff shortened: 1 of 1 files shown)


lazyfrosch:
Tested it on Debian unstable with Gnome 3.36 in Wayland.

Not really sure why this was added in the past...

Issue #26, opened 2020-04-19
Improve AppStream Metadata


lazyfrosch:
* https://appstream.debian.org/sid/main/issues/terminator.html
* https://salsa.debian.org/python-team/applications/terminator/-/blob/debian/master/debian/patches/appdata_improvement.patch
* https://salsa.debian.org/python-team/applications/terminator/-/blob/debian/master/debian/patches/appdata_in_metainfo.patch

stinnnnn:
Is there a problem resolving this issue?

buhtz:
Hope it is OK that I don't open a new issue. On Debian (I am just a user) I discovered some problems with your appstream metadata.

```
appstreamcli validate termina*xml
W: terminator.desktop:3: cid-desktopapp-is-not-rdns terminator.desktop
I: terminator.desktop:148: description-first-para-too-short The robot future of terminals
I: terminator.desktop:521: unknown-tag updatecontact
I: terminator.desktop:~: desktop-app-launchable-omitted
I: terminator.desktop:~: content-rating-missing
```

The primary problem is non-rdns `<id>`. In your case it usually would be `io.github.gnome-terminator.terminator`. An alternative is `io.github.terminator`. Important is that the first two commponents are derived from a real existing URL. It don't have to be Microsoft github. This decission is up to you as a maintainer.

The content-rating-missing can be fixed with this "empty" content rating.

```
    <!-- https://hughsie.github.io/oars/generate.html -->
    <content_rating type="oars-1.0" />
```

Your `updatecontact` is just a typo. Use `<update_contact>` ([docu](https://freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-update_contact)).

Add a `<launchable>` tag. e.g.

    <launchable type="desktop-id">terminator.desktop</launchable>

PR #53, opened 2020-05-01
Install AppStream data as .metainfo.xml

DarthGandalf:
Following advice from https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html


DarthGandalf:
Heh, I've noticed #26 now

PR #59, opened 2020-05-06
Fix layout launcher error

FernandoBasso:
The layout launcher crashing with an error upon attempting to select a layout.

```
$ ./terminator -s
```

Terminator-Session-Launcher-Small

Then select a layout, and this crash happens:

```
  File "/home/fernando/Projects/terminator/terminatorlib/layoutlauncher.py", line 87, in on_row_activated
	self.launch_layout()
  File "/home/fernando/Projects/terminator/terminatorlib/layoutlauncher.py", line 100, in launch_layout
	spawn_new_terminator(self.terminator.origcwd, ['-u', '-l', layout])
  File "/home/fernando/Projects/terminator/terminatorlib/util.py", line 335, in spawn_new_terminator
	cmd = os.path.join (cwd, sys.argv[0])
  File "/usr/lib/python3.8/posixpath.py", line 76, in join
	a = os.fspath(a)
TypeError: expected str, bytes or os.PathLike object, not NoneType
```

‘LayoutLauncher’ needs to know the current working directory, which as being passed as ‘None’ to `spawn_new_terminator()` in the `terminatorlib/layoutlauncher.py` . This commit fixes it by setting the current directory unconditionally so it is always available throughout the application.

**NOTE**: It only happens when invoking the layout launcher with the `-s` option as shown above. It works fine if we open the layout launcher with `Alt-l` _after_ Terminator is already running.

### DIFF
--- a/terminator
+++ b/terminator
@@ -68,6 +68,9 @@ if __name__ == '__main__':
   
     OPTIONS = terminatorlib.optionparse.parse_options()
 
+    TERMINATOR = Terminator()
+    TERMINATOR.set_origcwd(ORIGCWD)
+
     if OPTIONS.select:
         # launch gui, return selection
         LAYOUTLAUNCHER=LayoutLauncher()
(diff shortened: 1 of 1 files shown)


mattrose:
Taking a look at this now.

PR #60, opened 2020-05-06
Make LayoutLauncher window a little larger by default

FernandoBasso:
The Layout Launcher initial window is too small, especially the height.

Terminator-Session-Launcher-Small

This commit makes it a little larger by default:

Terminator-Session-Launcher-Larger


mattrose:
This looks much better thank you.

PR #61, opened 2020-05-07
Propagate tab-swictch events if there is only one tab

blackm0re:
If there is only one active tab, terminator should not consume tab-select events.
Those should instead be propagated, so they can be used by other programs "further down the line". Tab switch keybindings are often shared with terminal programs like WeeChat and irssi that also implement tabs.

### DIFF
--- a/terminatorlib/terminal.py
+++ b/terminatorlib/terminal.py
@@ -885,6 +885,20 @@ class Terminal(Gtk.VBox):
         # FIXME: Does keybindings really want to live in Terminator()?
         mapping = self.terminator.keybindings.lookup(event)
 
+        # Just propagate tab-swictch events if there is only one tab
+        if (
+                mapping and (
+                    mapping.startswith('switch_to_tab') or
+                    mapping in ('next_tab', 'prev_tab')
+                )
+        ):
+            window = self.get_toplevel()
+            child = window.get_children()[0]
+            if isinstance(child, Terminal):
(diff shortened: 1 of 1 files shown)


mattrose:
I'll take a look at this but any way you have of testing this change, even if it's just a description of how you use a common program, that would make it much easier for me to make sure it works as planned

blackm0re:
Hi and thanks for replying!

I will try to describe a normal use-case:
 - In Preferences -> Keybindings you map the names 'switch_to_tab_1', 'switch_to_tab_2' ... to respectively Alt+1, Alt+2...
 - Now when opening 2 or more tabs.. pressing Alt+2 sends you to the 2nd tab... etc
 - When opening one tab and starting a terminal program like irssii... pressing Alt+2 is not propagated (irssi will not switch to its 2nd tab, because the event doesn't reach it). Now in order to be able to use irssii with terminator, one has to have a different set of keybindings - one for terminator and one for irssii... since they are pointless in terminator unless terminator has multiple tabs, they can be reused using that small patch: multiple tabs - use the bindings, one tab - propagate the event.

I hope that this brings some clarity. I have been using this patch for almost two years now.

Regards

mattrose:
Cool, thanks for responding.  Hopefully I'll take a look at it this weekend.

mattrose:
Sorry I haven't looked at this until now, but I did, and it all looks good.  Thanks for the patch, and thanks for your patience

PR #67, opened 2020-05-09
Update AUTHORS with a full list from GIT history

lazyfrosch:
This is an approach to mention everyone contributing to Terminator.


lazyfrosch:
Anyone thoughts on this? I like to do this on my projects, and it can be easily refreshed for releases.

Issue #65, opened 2020-05-09
terminator runs in network namespace only with -u option (dbus)


lazyfrosch:
Copied from Launchpad: https://bugs.launchpad.net/terminator/+bug/1874722

Terminator cannot be started inside a network namespace.

Running terminator 1.91 on Debian 10 (buster), Kernel 4.19.0-8-amd64 x86_64, Xfce 4.12.4

I generally setup the new network namespace (call it nns1), then I issue:

```
$ sudo ip netns exec nns1 su -c "terminator -d -d " mario

(terminator:16069): dbind-WARNING **: 14:40:12.594: Couldn't connect to accessibility bus: Failed to connect to socket /tmp/dbus-HMTnanv0jP: Connection refused
ConfigBase::__init__: Borg::__init__: Preparing borg state for ConfigBase (/usr/share/terminator/terminatorlib/borg.py:47)
noclass::get_config_dir: Found config dir: /home/mario/.config (/usr/share/terminator/terminatorlib/util.py:194)
ConfigBase::load: looking for config file: /home/mario/.config/terminator/config (/usr/share/terminator/terminatorlib/config.py:604)
ConfigBase::load: config validated successfully (/usr/share/terminator/terminatorlib/config.py:632)
ConfigBase::load: ConfigBase::load: Processing section: global_config (/usr/share/terminator/terminatorlib/config.py:635)
ConfigBase::load: ConfigBase::load: Processing section: keybindings (/usr/share/terminator/terminatorlib/config.py:635)
ConfigBase::load: ConfigBase::load: Processing keybindings: zoom_in (/usr/share/terminator/terminatorlib/config.py:664)
ConfigBase::load: ConfigBase::load: Processing keybindings: zoom_out (/usr/share/terminator/terminatorlib/config.py:664)
ConfigBase::load: ConfigBase::load: Processing keybindings: zoom_normal (/usr/share/terminator/terminatorlib/config.py:664)
ConfigBase::load: ConfigBase::load: Processing keybindings: new_tab (/usr/share/terminator/terminatorlib/config.py:664)
ConfigBase::load: ConfigBase::load: Processing keybindings: cycle_next (/usr/share/terminator/terminatorlib/config.py:664)
ConfigBase::load: ConfigBase::load: Processing keybindings: cycle_prev (/usr/share/terminator/terminatorlib/config.py:664)
[...]

PR #88, opened 2020-05-16
Do not crash when dbus server is unavailable, just emit an error message

mattrose:
In a situation where we have all of the dbus libraries installed, but there is no dbus server available, for whatever reason (see bug #65 for one), don't crash, just proceed without dbus

This is actually a pretty rare situation, but I can see this happening under some unusual setups.  It happened to me running on a mac.  I got around it by installing the dbus server, but some people may not want to do that.

### DIFF
--- a/terminatorlib/ipc.py
+++ b/terminatorlib/ipc.py
@@ -11,7 +11,7 @@ from .borg import Borg
 from .terminator import Terminator
 from .config import Config
 from .factory import Factory
-from .util import dbg,  enumerate_descendants
+from .util import dbg, err, enumerate_descendants
 
 CONFIG = Config()
 if not CONFIG['dbus']:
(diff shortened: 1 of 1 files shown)


r2rien:
When I submitted a patch 
(based on initial work from Kacper Wysocki)
for opening a new tab in an already running instance of another python application: meld,
This is how it ended, as kai (meld author/maintainer) wanted also to never crash when handling or not dbus:  
https://gitlab.gnome.org/GNOME/meld/-/commit/e08baf625ba002e1c6f6ea46d9222bc3e1ec289a

my 2cts

mattrose:
@r2rien that is probably better, but I think that my patch does enough.  The only situation I can see possibly causing a problem is if the dbus server is running or reachable when terminator starts, and then stops or becomes unreachable while terminator is running.  

However, testing this on macos or FreeBSD did not cause a crash and on later versions of both Fedora and Ubuntu if you kill the dbus server the whole desktop crashes.

PR #92, opened 2020-05-17
WIP/crazy idea: add feature to embed browser tab

terceiro:
Here is a crazy idea.

I need to do a remote presentation/live demo in a few days, and I anticipate that I will need to switch back forth between a terminal and a browser. I don't want to share my entire desktop, and I don't want to be constantly picking a window to share.

So I wrote this to be able to embed a simple WebKit2 based browser tab in terminator.
It's not finished, but it kind of works for me so far.

Is this crazy? Would people be willing to carry this feature going forward? I admit it's very niche, but it's also very useful for the use case it's intended to cover.

### DIFF
--- a/terminatorlib/browser.py
+++ b/terminatorlib/browser.py
@@ -1,0 +1,156 @@
+"""A simple browser widget with WebKit2"""
+
+import gi
+gi.require_version('WebKit2', '4.0')
+from gi.repository import GObject, Gtk, WebKit2
+from .terminator import Terminator
+
+class Browser(Gtk.VBox):
+    __gsignals__ = {
+        'title-change': (GObject.SignalFlags.RUN_LAST, None,
+            (GObject.TYPE_STRING,)),
+        'tab-new': (GObject.SignalFlags.RUN_LAST, None,
+            (GObject.TYPE_BOOLEAN, GObject.TYPE_OBJECT)),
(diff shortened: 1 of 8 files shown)


mattrose:
That is absolutely crazy.  I also kinda love it :). I'm gonna play with it a bit myself.

lazyfrosch:
Not sure what to say... :laughing:

mattrose:
Now I kinda want to see if I can run it in it's own pane.

terceiro:
> Now I kinda want to see if I can run it in it's own pane.

that would be interesting. I had an even crazier idea, which was to be able to have "arbitrary" gtk widgets in place of terminals in any layout, but I think that's probably too far. :roll_eyes:

Sarke:
Wouldn't it have been easier to just use a web-based console in one browser tab?

mattrose:
I had a crazy thought last night.  I haven't looked at the code, but it might be possible to restructure this as a plugin.  That way it can be used and possibly even distributed like some other plugins, without adding a lot of complexity or extra requirements.

lazyfrosch:
So should we really pursue this? How stable and safe is the GTK WebKit lib for us to use?

I'm afraid of supporting any browser code :laughing:

terceiro:
nah

PR #97, opened 2020-05-18
Improvements to data/terminator.desktop

dmaphy:
I tried building an RPM from latest master and stumbled over some issues when running the desktop-file-install command for `data/terminator.desktop`:

```
+ desktop-file-install --dir=/home/dmaphy/rpmbuild/BUILDROOT/terminator-2.0-0.1.ec2d50cdgit.fc32.x86_64/usr/share/applications data/terminator.desktop
/home/dmaphy/rpmbuild/BUILDROOT/terminator-2.0-0.1.ec2d50cdgit.fc32.x86_64/usr/share/applications/terminator.desktop: error: file contains key "_Name" in group "Desktop Entry", but key names must contain only the characters A-Za-z0-9- (they may have a "[LOCALE]" postfix)
/home/dmaphy/rpmbuild/BUILDROOT/terminator-2.0-0.1.ec2d50cdgit.fc32.x86_64/usr/share/applications/terminator.desktop: error: file contains key "_Comment" in group "Desktop Entry", but key names must contain only the characters A-Za-z0-9- (they may have a "[LOCALE]" postfix)
/home/dmaphy/rpmbuild/BUILDROOT/terminator-2.0-0.1.ec2d50cdgit.fc32.x86_64/usr/share/applications/terminator.desktop: error: required key "Name" in group "Desktop Entry" is not present
Error on file "data/terminator.desktop": Failed to validate the created desktop file
```

Thus I've tried fixing these here plus I added the change we apply since years in the Fedora package (of which I am not sure if we can apply it generally):

```
-[NewWindow Shortcut Group]
+[X-NewWindow Shortcut Group]
```


mattrose:
I think the underscores("_") in front of `Name` and `Comment` are necessary for the intltool-merge stuff for localization.  Can you put the underscores back?

dmaphy:
Yep. I've put those underscores back where they belong. :-)

mattrose:
@lazyfrosch it looks to me like the following patch fixes things for Fedora, but Debian and Ubuntu work without it.  At least when I look at the package in Ubuntu 20.04 I see: `[NewWindow Shortcut Group]` not `[X-NewWindow Shortcut Group]`, and I know when building on Fedora, it doesn't work without the patch that @dmaphy is submitting.

PR #104, opened 2020-05-22
Update pt_BR.po

chclxds:


mattrose:
Thanks for this. @lazyfrosch is just in the middle of redoing the localization scripts again, so we'll definitely pull this in when we're done.

PR #110, opened 2020-06-02
add debug logging to searchbar

mattrose:


mattrose:
This would have helped with the search stuff not working like in #4.  I notice that's been fixed but I may pull it in anyways, as there is currently no debug logging in the searchbar code at all.

PR #112, opened 2020-06-02
fix traceback when closing a window

mattrose:


mattrose:
Basically, if you open a window, then open a tab in that window, and then open a new window using the `terminator` command, and then close the first window you opened with the two tabs in it, then the call to get_children()[0] will fail with an IndexError, because the window won't have any children.

PR #124, opened 2020-06-11
add line_height config variable

mattrose:
This is a WIP right now.  I'm going to wire up the preferences dialog next.  If you want to try this out, you can add a `line_height` entry to a profile in `~/.config/terminator/config`


srt0:
@mattrose It works. :)

Screenshot

PR #126, opened 2020-06-11
Adding maximize option

### LINKED ISSUE #125: align maximize option spelling with gnome-terminal
terminator has the maximize option spelled differently from gnome-terminal (and xfce-terminal)

maximise (UK) versus maximize (US)

```
terminator --maximise

versus

gnome-terminal --maximize
xfce-terminal --maximize
```

As a drop-in replacement for gnome-terminal it would be nice if the most used cli options were somewhat in sync.

As terminator is already released with --maximise option one solution could be to recognoze both --maximise and --maximize options ?

qckzr:
This PR fixes #125 

@mattrose let me know if this works, I also modified the doc section.

### DIFF
--- a/doc/terminator.1
+++ b/doc/terminator.1
@@ -25,6 +25,8 @@ Show the version of the Terminator installation
 .TP
 .B \-m, \-\-maximise
 Start with a maximised window
+.B \-M, \-\-maximize
+Start with a maximized window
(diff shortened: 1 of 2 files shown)


mattrose:
YOU ARE AWESOME! I will check this out right now.

PR #130, opened 2020-06-14
Allow Alt + mouse wheel up/down to scroll apps

mckellyln:
Hi,
Thanks to all for Terminator!
Here is a PR for the small change discussed before in:
https://github.com/gnome-terminator/terminator/pull/46
to allow Alt + mouse wheel up/down to scroll.
I wanted to make it so I could check for Alt + wheel in application and scroll 2x/4x more but that is still to be determined how.  This at least gets us to scroll normally.
thx,
-m


mattrose:
Thanks for the contribution.  I haven't forgotten about you, it's just that my current "setup" (as the kids would say) is sitting on my couch with my laptop, and I haven't dug out my mouse recently to test your changes.  I'll test and merge as soon as I can.

mckellyln:
No problem :-)

terminatorlib/terminal.py
```diff
@@ -1018,6 +1018,9 @@ def on_mousewheel(self, widget, event):
             elif event.direction == Gdk.ScrollDirection.DOWN or SMOOTH_SCROLL_DOWN:
                 self.scroll_by_page(1)
                 return (True)
+        if event.state & Gdk.ModifierType.MOD1_MASK == Gdk.ModifierType.MOD1_MASK:
```
GrzegorzDerdak, line 1021: `Gdk.ModifierType.MOD1_MASK == Gdk.ModifierType.MOD1_MASK` would always evaluate to true so it can be removed from this conditional

mckellyln: ok, sure.  I was copying the coding style of the other event.state conditionals near by.

GrzegorzDerdak: Changing it to `!= 0` would be better. Also, I just notice it, please use `and` instead of `&` - it's more pythonic. So the full conditional could be just `if event.state and Gdk.ModifierType.MOD1_MASK != 0`.

mckellyln: Is 'and' bitwise ?  Not sure how it can be both bitwise and logical ?  I think its pretty common to use '&' for bitwise ?
Would parens help ?
Again I'm just following the same style as several other lines above it.

GrzegorzDerdak: Ok, my mistake. Maybe just leave as it is since other lines are in the same style. Sorry for the confusion.

mckellyln:
Closing as its probably too far behind master now and its such a small change.
thx,
-m

PR #138, opened 2020-06-19
Solved a bug in prefseditor.py

robertoetcheverryr:


mattrose:
Can you rebase the commit so that it doesn't contain the merge commit?

PR #144, opened 2020-06-22
issue 44, add option for case sensitive search

dugb:
Added regex flags to allow for caseless search
Added a Match Case checkbox to the search bar
Cleaned up a couple deprecation warnings in searchbar.py

### DIFF
--- a/terminatorlib/regex.py
+++ b/terminatorlib/regex.py
@@ -11,6 +11,10 @@ from gi.repository import GLib, Vte
 # the corresponding bits are defined here:
 #   https://vcs.pcre.org/pcre2/code/trunk/src/pcre2.h.in?view=markup
 PCRE2_MULTILINE = 0x00000400
+PCRE2_CASELESS = 0x00000008
+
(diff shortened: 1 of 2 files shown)


mattrose:
This looks good I'm just going to run a few checks first

mattrose:
Found one traceback while running this under Python 2.7
```
  File "/home/mattrose/terminator/terminatorlib/searchbar.py", line 122, in match_case_toggled
    self.regex_flags_pcre2 = (regex.FLAGS_PCRE2 | regex.PCRE2_CASELESS)
TypeError: unsupported operand type(s) for |: 'NoneType' and 'int'
```
Is there a different syntax for setting flags in Python2 than python3?

dugb:
Gotcha.

I'm thinking that issue should also pop up with python3.

regex.FLAGS_PCRE2 will be `None`, if the system does not have PCRE2 regex support.  in which case it would fall back to using the GLIB regex.

I'll add a try/except to catch the TypeError.

mattrose:
Awesome, thank you.  I think you're right.

mattrose:
Thanks @dugb this is great!

PR #150, opened 2020-06-27
Fix: Allow Key Bindings with Shift-Modified Keys

### LINKED ISSUE #149: Key Bindigs That Contain a Key Modified by a Shift Key Don't Work
The issue was already reported here: https://bugs.launchpad.net/terminator/+bug/999897.
If a key binding contains a key that was modified by a Shift key, think [ and {, such a key binding won't work at all.

**Steps to reproduce:**

1. Go to Preferences -> Keybindings.
2. Double click on a current key binding next to 'Toggle fullscreen' or any other action.
3. Press `Ctrl + Shift + 1`, that is hold Ctrl, Shift and then press 1.
4. The resulting key binding is now set to `Shift + Ctrl + !` instead of `Ctrl + !`.
5. Go to the main terminal window and press `Ctrl + Shift + 1` to trigger the action — the key binding won't work!

dkmvs:
This PR should fix #149. 

The idea is to check whether a key value will change if a Shift key is down (see [level](https://developer.gnome.org/pygtk/stable/class-gdkkeymap.html)). If it will, then remove the Shift modifier from `mods`. 

One exception: if a key binding contains a letter then the Shift modifier is not removed. This is because, for some reason, a key value of a letter is never modified by the Shift modifier and always corresponds to a key value of a lowercase character. This is already handled in `terminatorlib/keybindings.py`. 

I'm new to GTK, so not quite sure how to emulate signals with PyTest. I tested the following key bindings (manually) to make sure it works: 

- `Ctrl + Shift + 1` translates to `Ctrl + !`
- `Ctrl + A` translates to `Ctrl + A`
- `Ctrl + Shift + A` translates to `Shift + Ctrl + A`
- `Ctrl + Shift + F1` translates to `Shift + Ctrl + F1`
- `Ctrl + Shift + Page Up` translates to `Shift + Ctrl + Page Up`
- `Ctrl + Shift + Up` translates to `Shift + Ctrl + Up`
- `Shift + Up` translates to `Shift + Up`
- `Shift + A` translates to `Shift + A`
- `Ctrl + Shift + [` translates to `Ctrl + {`

### DIFF
--- a/terminatorlib/prefseditor.py
+++ b/terminatorlib/prefseditor.py
@@ -1550,6 +1550,21 @@ class PrefsEditor:
 
     def on_cellrenderer_accel_edited(self, liststore, path, key, mods, _code):
         """Handle an edited keybinding"""
+        if mods & Gdk.ModifierType.SHIFT_MASK:
+            key_with_shift = Gdk.Keymap.translate_keyboard_state(
+                self.keybindings.keymap,
+                hardware_keycode=_code,
+                state=Gdk.ModifierType.SHIFT_MASK,
+                group=0,
+            )
+            keyval_lower, keyval_upper = Gdk.keyval_convert_case(key)
+
+            # Remove the Shift modifier from `mods` if a new key binding doesn't
+            # contain a letter and its key value (`key`) can't be modified by a
+            # Shift key.
+            if key_with_shift.level != 0 and keyval_lower == keyval_upper:
+                mods = Gdk.ModifierType(mods & ~Gdk.ModifierType.SHIFT_MASK)
+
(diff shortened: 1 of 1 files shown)


mattrose:
Thank you so much for this!   I recall wanting to move over the original LP issue but I missed it.  I'll review it within the next day or so and get back to you if I have any problems.  If not, I'll just merge it in as is. I'm not sure how to emulate signals with pytest either, so manual testing is definitely the way to go, and thanks for testing it out as well.

mattrose:
fixes #149

mattrose:
Thanks again for your help

Issue #129, opened 2020-06-13
Custom color palette not being memorized


SCCOBN:
This is an easily reproducible bug, click preferences, profiles, colors, configure custom color palette. I like colors from manjaro distro on my arch. Then close everything. Next time you visit same location in preferences, you will see that your color scheme will be Linux, instead of custom. 

However, if you select custom again, without configuring each color, you will have your colors that you configured earlier. 

In any case, this is an annoying bug and bugs me cuz every time I have to go into preferences and configure something, I have to change the platte to custom in order to have my colors.

mattrose:
Looked into this a bit.  The prefs pane appears to save the palette in `.config/terminator/config` just fine.  I think the problem is that the preference pane doesn't ever reload the new config file until you restart.  Can you try reproducing the problem, except, after you close everything, make sure there are no terminator instances running?

dugb:
I think what is going on here is that a users 'custom palette' is not saved in .config/terminator/config as a custom-palette. It is saved as `palette`.

So if a user sets up all their colors for a custom palette, that palette is saved as `palette`.

But then if they decide to try out Tango palette the Tango palette is saved as `palette` and overwrites all their custom colors.

Then if the user decided they don't want to use Tango palette anymore and they want to go back to using the custom palette they setup, all their custom colors are gone.

I recommend continue saving the current palette as `palette` in the config and adding `custom-palette` to save the users custom colors.

I'll look into make changes.

mattrose:
@SCCOBN PR #160 appears to fix this, can you confirm?

PR #160, opened 2020-07-08
Allow live previewing of profile color changes

dafrito:
This pull request allows live previewing of profile color changes. Specifically, the foreground, background, and palette colors, when changed, will be shown immediately in the terminator window. Prior to this pull request, the preferences would need to be saved before changes were shown in the window. This allows a more intuitive experience of changing colors.

### DIFF
--- a/terminatorlib/preferences.glade
+++ b/terminatorlib/preferences.glade
@@ -2312,7 +2312,7 @@
                                         <property name="label" translatable="yes">_Text color:</property>
                                         <property name="use_underline">True</property>
                                         <property name="justify">center</property>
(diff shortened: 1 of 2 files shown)


mattrose:
Thank you so much for this.  Fernando was asking about this very thing on gitter.im this morning.

mattrose:
fixes #129 

Thanks again for doing this.  This makes a really good base for further changes to the preferences that I had in mind

mattrose:
@dafrito would you mind doing another PR for live previewing the titlebar background and foreground colors?

PR #170, opened 2020-07-22
confirm window close when child processes exist

mattrose:
This just asks the user to confirm that they want to close the window when there is a child process running in the window that is attached to it.


mattrose:
I was going to try and jam this into the existing "Close multiple terminal" dialog, but then I noticed how tilix did it, and I decided to rethink the entire thing.  I'm gonna abandon this line of work but definitely fix the issue

PR #181, opened 2020-08-14
Format shortcuts

aadrian:
Format shortcuts a little in the README

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,11 +1,11 @@
 Terminator
 ==========
 
-by Chris Jones <cmsj@tenshu.net> and others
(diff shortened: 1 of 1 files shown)


mattrose:
Can you make sure that the trailing spaces are preserved in the README so that line breaks still exist?

aadrian:
@mattrose not sure what you mean: the line breaks are there. 

I just formatted it so that it is more readable - see here the rendering : https://github.com/aadrian/terminator/tree/patch-1 

Also for readability, where there's one shortcut that's short than it's inline,  else they are as bullet list.

mattrose:
Cool, looks good.

PR #182, opened 2020-08-15
Add a 'title bar at bottom' option

phidebian:
Hi All,
This is my first attempt to colaborate to an open source project. I may have goofed in many ways.

This pull request is a demonstrator of a feature I need, dunno if it would be popular or not.

In the config pref menu I add a Title bar at bottom check, once checked, exist terminator then run it again, then the tittle bar is at bottom. 

In my use case I stick in there my current dir along with my git branch, and then my prompt is simply '$'  (or '#' if I am root logged), I use this on big screen, more that 100 lines, having the $PWD and git BR on the bottom close to my prompt (where I type all the time) is very convinient, having this on top make a lot of tiring eyes movments.

Self crtitics of the implementation. First of all bear in mind I am python ilitereate.

terminator/remotinator, I hardcoded python3 in the #! (shebang) because the current python point to python2 on my system, I use virtualbox which despotically install the python-is-python2 package. Since this version of terminator is python3 dependent (doesn't run with python2) I don't see why setting  #!/usr/bin/python not knowing if we will land in python 2 or 3.
I may be completly off base here, and may be there are good reasons to keep /usr/bin/python, but to me it looks like we will bump into problems again and again, since there is only 2 such file with #! it looks to me a good things to do to make it python3. If one day python4 come to the dance and if we port to it, it would simply be a matter of fixing 2 files...

The rest of the code seems to work to me except 1 things I don't have the knowledge to fix, that is when toggling the state of 'Title bar at bottom' check box it require an exit of terminator and re-run to take effect, may be someone can do the last mile and add a kinda layout refresh when the layout changed.

Last thing, I  patched the .glade file manually may be glade  should be used...

I like terminator very much, I did try an whole lot of alternatives (tilix, etc..) I crossed out the one who can't scroll windows like tmux. I got many more things to do on terminator but will propose them one at a time.

Cheers,
Phi

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,11 +1,11 @@
 Terminator
 ==========
 
-by Chris Jones <cmsj@tenshu.net> and others
+by Chris Jones <cmsj@tenshu.net> and others.
 
 ## New home on GitHub
 
-In April of 2020 we started moving Terminator to GitHub. A new team wanted to continue the work of the original authors.
+In April of 2020 we started moving **`Terminator`** to GitHub. A new team wanted to continue the work of the original authors.
 
 Unfortunately we are not able to adopt the Launchpad project, so we could only inform users were possible. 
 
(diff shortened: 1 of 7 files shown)


mattrose:
Looks good.  Thanks so much.

PR #186, opened 2020-08-19
Apparently mint does not like it when you call unmaximize twice

### LINKED ISSUE #168: 'maximized' config property is misspelled and prevents position from working.
LM 20
terminator 1.91

1. Launch Terminator and move the window where you would like it to be. 
2. Open Preferences (right+click) and choose Layouts and then Window0. Click Save
3. Relaunch

Notice that in the Preferences file `~/.config/terminator/config` the key `maximised` is wrong and should be `maximized`. 

Prevents `position`  window positioning from working. Probably crashes.  After I fix spelling to `maximized` window positioning works as expected.

Here is my misspelled version of config after I saved Preferences.

```
[global_config]
  geometry_hinting = True
  title_transmit_bg_color = "#4e9a06"
[keybindings]
  paste = <Primary>v
[profiles]
  [[default]]
    cursor_color = "#aaaaaa"
    scrollback_lines = 1000
    copy_on_selection = True
[layouts]
  [[default]]
    [[[child0]]]
      type = Window
      parent = ""
      order = 0
      position = 262:62
      maximised = False   <------------- THIS LINE IS MISSPELLED  should be 'maximized'
      fullscreen = False
      size = 939, 626
      title = ~
      last_active_term = 05ec768e-6ece-4f28-80c9-7a08f780676f
      last_active_window = True
    [[[terminal1]]]
      type = Terminal
      parent = child0
      order = 0
      profile = default
      uuid = 05ec768e-6ece-4f28-80c9-7a08f780676f
[plugins]
```

mattrose:
This fixes #168 and I tried to very gingerly fix it, but I kinda want to test the crap out of this before merging, as the current code works everywhere but Mint.

### DIFF
--- a/terminatorlib/window.py
+++ b/terminatorlib/window.py
@@ -344,7 +344,9 @@ class Window(Container, Gtk.Window):
         if value == True:
             self.maximize()
         else:
-            self.unmaximize()
+            ### Apparently if you call unmaximize twice in Mint it breaks other stuff
+            if self.props.is_maximized == True:
+                self.unmaximize()
 
     def set_fullscreen(self, value):
         """Set the fullscreen state of the window from the supplied value"""


mattrose:
user plans to open a Mint bug instead, as this only affects Mint

PR #189, opened 2020-08-22
Use the term 'zero padded' instead of 'padded'.

phidebian:
Use the term 'zero padded' instead of 'padded' in prefs key binding and terminal pop up menu, to be consistent with the doc.
https://gnome-terminator.readthedocs.io/en/latest/gettingstarted.html?highlight=padded
Beside it will be consistent with the docco, it also took me a while to understand what 'padded' mean, specially as long as you don't open more then 10 terminals where it looks like the same as non padded.

OTOH  I still don't understand what is hte use case of terminal id is for so may be this is not important.


phidebian:
Thanx ;-)

Issue #156, opened 2020-07-02
Feature Request: Key binding option for creating group in current tab


bboytian:
I find the "split to this group" option extremely useful when you have the "broadcast to group" option activated. This allows for creating multiple terminals and running the same commands on the fly. However, to create a group, I have to use my mouse and click "New group...". Could this be set to a key binding?

mattrose:
Don't see why not...  I'll look into this.

PR #195, opened 2020-08-25
Key binding option for creating group in current terminal

mattrose:
This behaves exactly as the "New Group" titlebar menu option.  It will move the focus to the randomly selected group name, and let you edit it, and then returns focus.

It's currently disabled, but settable in the preferences.


mattrose:
this fixes #156

Issue #159, opened 2020-07-06
Cannot "Drag and Drop" Folders or Files


Dhamma-valence:
Expected behaviour:
Drag a file/folder icon and drop it onto Terminator to be able to enter the directory path in Terminator's buffer.

Steps to reproduce:
Open terminator in terminal (meta) in order to get the debug output. Then open any file browser and attempt to drag the icon and drop it into Terminator. Get the following error message in the terminal you opened terminator in (still meta).

`TypeError: a bytes-like object is required, not 'str'
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/terminatorlib/terminal.py", line 1122, in on_drag_data_received
    txt = txt.split('\n')[0]
TypeError: a bytes-like object is required, not 'str' `

dugb:
I can not reproduce this using the latest from master.

Looking at the code, terminal.py lines 1142 and 1145 the decoding from str to bytes is being done.

I can reproduce this using terminator from the debian buster repository, version 1.91

What version of terminator are you using?

mattrose:
This was fixed a while back in master. see #54 for fix

Dhamma-valence:
Yes thank you, I build from the master instead of using Debian Stable repos and it works!

My only issue with it now is that dragging and dropping a file appends `file://` to the beginning of the directory string, which results in Terminator outputting `"No such file or directory"`. 

I'm unsure if this is on the OS or Terminator's side, but do you know of a fix / way to remove or prevent this behaviour?

mattrose:
hmm, shouldn't be doing that. Lemme check

mattrose:
@Dhamma-valence I was unable to reproduce this.  when I drag a file from the Files app in GNOME it just prints out the file name in quotes, not a file url.  There is a lot of weird complicated logic around so it's possible that it's not doing the correct the thing in your case.  Can you tell me what your desktop environment or window manager is, and what program you're dragging the file from?

im-AMS:
Hey,
I had the same issue of `file://`
I run manjaro KDE,

I think I have managed to fix it, ill make a pull request, please check it out

Thank you

Dhamma-valence:
I use KDE, with the default Kwin WM on Debian Testing -
 Krohnkite Kwin tiling script too but that's likely irrelevant.

I _think_ this might be related to another error I was having with something called KIOExec on KDE but I could be wrong. 

[...]

PR #197, opened 2020-08-26
drag and drop error #159

im-AMS:
since I had the same issue #159 , on manjaro KDE I have tried to fix it.
please have a look at it, and see if it works for all the users.

Thank You😊


mattrose:
So, I think I figured it out.  I think, Manjaro KDE is sending file URIs in the format `file://file_name` and the proper URI format is `file:///file_name` 

In the spirit of enforcing RFCs, I would open a bug on the Manjaro KDE file manager telling them to format the file string data correctly.

In the spirit of "Be conservative in what you send and liberal in what you accept", if you can fix the `on_drag_data_received` method, rather than the `feed` method, I'll consider merging that.

PR #198, opened 2020-08-27
As per suggestion

im-AMS:
I have tried fixing it, I dint want to tamper with the existing code, the solutions seems lazy but it does work for me, and hopefully for all users.😊


mattrose:
Sorry I wasn't able to use this.  There was some deeper stuff going on, so I had to take a different approach

im-AMS:
no worries!... it was a pleasure doing something for the FOSS community!

ps it was my 1st pull request ever 😊

PR #211, opened 2020-09-18
Fix: Allow `Shift+Tab` Key Binding

dkmvs:
PR #150 introduced a bug which made it impossible to add the `Shift+Tab` key binding in `Preferences>Keybindings`. This PR should fix this. 

The commit messages in this PR have a more detailed explanation about the nature of the bug. 

I also added automated tests to check that editing a key binding in `Preferences>Keybindings` with the predefined key combinations, including `Shift+Tab`, produce the expected accelerator.


mattrose:
Thanks for this!  I'll check it out tomorrow.

mattrose:
Sorry it took me a while.  It looks good.

Issue #213, opened 2020-09-18
New feature: A Separate Json file for dynamic Layouts


dvdlevanon:
Adding the following command line arguments 
```
--layout path/to/layout.json
```

The Json file may provide an easy way to configure a collection of tabs and terminals using common pre-defined layouts, and also may support basic theming like background and text color.

This would allow to:
1. Edit Layouts using a Text Editor instead of the "Preferences > Layouts" dialog.
2. Share layouts easily between machines and people.
3. A way to keep layouts under source control (git).

-----

For years I used terminator and this feature was missing. so I wrote my own script that takes a list JSON files, representing layouts and profiles, and build a unified INI file for the terminator. Those files are also in the source control so my colleagues can use them easily. 

Now that I see terminator is under active development again, I think of adding this feature to the code base. Would like to get your feedback and suggestions.

mattrose:
This sounds interesting.  I think it might be better to have profiles defined in JSON rather than layouts.  Profiles have the functionality that you need (judging from your description), and can contain layouts as well.  

Personally, I'd be very comfortable working on this with you as well, just because it would be a nice change from all the GUI programming I've been doing lately with trying to get background image support back, and definitely an area in which I'm more comfortable :)

dvdlevanon:
Hi,

Thanks for your response @mattrose, would be glad to work with you, I already started something and will have it ready for push soon, will share the PR.

Regarding layout vs profiles, I might miss something, currently a `.config` file looks like:
```[profiles]
  [[Monitor]]
    allow_bold = False
    icon_bell = False
    background_color = "#100000"
    background_darkness = 0.93
    background_type = transparent
    font = Monospace 16
    foreground_color = "#DCDfff"
    show_titlebar = False
    scrollbar_position = hidden
    scrollback_lines = 3000
    scrollback_infinite = True
    use_system_font = False
[layouts]
  [[Monitor]]
    [[[root]]]
      fullscreen = False
      last_active_window = True
      maximised = True
      order = 0
      parent = ""
      position = 1920:0
      size = 1920, 1080
      title = Monitor
      type = Window
    [[[tabs]]]
      active_page = 0
      labels = htop, cmus
      order = 0
      parent = root
      type = Notebook
    [[[htop]]]
      command = htop ; bash
      order = A
      parent = tabs
      type = Terminal
    [[[cmus]]]
      command = cmus ; bash
      order = B
      parent = tabs
      type = Terminal
```

[...]

PR #214, opened 2020-09-19
layout file - initial commit - work in progress

dvdlevanon:


mattrose:
Sorry it took me a while to get to these, looking at them now.

mattrose:
fixes #213

mattrose:
So, it looks really good, but I loaded up 2-3-grid.json and seem to have gotten myself into a state where there is an [[__internal_json_layout__]] section in the config and when I load up terminal from the regular config I get this errror
```
[mattrose@sorrento terminator]$ python3 terminator 
ConfigBase::load: config format is not valid
[layouts,__internal_json_layout__,root]: parent is invalid
```

dvdlevanon:
Hi @mattrose 

It happened to me as well, I'm looking into it now and will update soon.

Thanks for the review

dvdlevanon:
The `__internal_json_layout__` and `__internal_json_profile__` is the internal "Fake" layout and profile created by the `--config-json` argument. 

I put them in the `config.base` to make them a first citizen layout/profile and avoiding special care across all the code base.

Unfortunately, the "Fake" layout and profile found their way to the global config file, via the `save` mechanism. Indirectly causing errors when running terminator in the regular mode.

In order to solve it, I excluded those special names from being saved to the global config file.
https://github.com/gnome-terminator/terminator/pull/214/commits/b66999618067c314d16e0d0f6d7cad2ed6dfe0f0

mattrose:
Thanks for looking at this.  I'll pull down the new code later this morning (I'm on EST in Canada) and play with it.

mattrose:
This looks really good.  It's well organized and only touches the core code where it has to.  Is there anything else you want to do before merging?

dvdlevanon:
Thanks.

Before merging, I would like to:
1. Remove the example files, they should be in some doc file but not in the repository.
2. Let it run on my PC for a few days before merging, I'm working with it on a daily basis and I want to make sure nothing harm due to this change - Lets say a week from now.
3. As I'm not familiar with the development, I would be glad to hear what other steps need to be taken.

mattrose:
1.  I think they should maybe stay in the repository, maybe move them to the data subdirectory.

2.  Definitely more soak time is always good.  I'll definitely use this more myself over the next week, just to get a feel for it.  I did notice when I was working on #217 I created new config entries, and when I ran your code, it re created the config file without them (likely because those config entries didn't exist in the code in your branch).  I'm not sure that this should be fixed, or whether it would be worth how much work it would take to fix it, because let's be honest, it would only happen to people developing terminator, but it's something to keep in mind.

3.  Unit tests would be nice, if you're comfortable writing them.  They can make sure nothing that's horribly broken sneaks in.  If you're not comfortable writing them I would definitely merge it in anyways, but more unit tests is one of my goals

dvdlevanon:
1. I'll move them to `data`
2. Once both PRs will be merged the issue will be gone, correct?
3. Yes unit tests would be nice, not my strong side though

dvdlevanon:
@mattrose I used it for a while, everything seems to be working fine.
I think we are good to merge now

mattrose:
This looks great.  I'm planning on releasing 2.0 tomorrow or the day after, do you mind if I wait until after I tag and release, before pulling this in?

dvdlevanon:
Sure fine by me, I'm using "my version" of terminator anyway. so whenever you want to pull it its fine

mattrose:
Thanks Again!

PR #217, opened 2020-09-24
Background image

### LINKED ISSUE #5: No option for background image in gtk3 version
I remember we could use a background image as terminal background. It would be great if we could have this option back, [launchpad #1635704](https://bugs.launchpad.net/terminator/+bug/1635704)

mattrose:
fixes #5 

This is a very barebones implementation so far, and doesn't even have GUI controls yet.  (Have I mentioned how much I hate Glade?) but...

Witth this code, in your profile, you can now set `background_image = $image_file` and it will set the background of the terminal image to the image.  This is the background for each pane, so the image will no span multiple panes, but instead will repeat (I think I can configure that if it's a popular feature request, but I don't want to get too ahead of myself)

You can also set `background_alpha = x` where x is a float between 0 and 1 (eg: 0.55) which will blend in the background color from your theme to aid readability.

I have to say, I was a huge skeptic on the usability of this, it just seemed like a popular feature and people did come up with various use-cases, but I really like the way this turned out.

### DIFF
--- a/terminatorlib/config.py
+++ b/terminatorlib/config.py
@@ -250,6 +250,8 @@ DEFAULTS = {
                 'autoclean_groups'      : True,
                 'http_proxy'            : '',
                 'ignore_hosts'          : ['localhost','127.0.0.0/8','*.local'],
+                'background_image'      : '',
+                'background_alpha'      : 0.0
             },
         },
         'layouts': {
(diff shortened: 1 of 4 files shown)


mattrose:
TODO:

add GUI controls in prefseditor.glade and wire them up

PR #223, opened 2020-10-01
WIP: initial work on #74

### LINKED ISSUE #74: Reimplement "unfocused dim" using overpainting
Currently Terminator dims unfocused panels by temporarily setting a different color palette.

This has the following drawbacks:
- Doesn't affect the 16M truecolors.
- Doesn't affect a palette color if it was overridden using the OSC 4 escape sequence, because in VTE the OSC values take precedence over the API ones.
- As per https://gitlab.gnome.org/GNOME/vte/-/issues/242, in a future VTE release it'll probably stop working for the 256-color palette too, and will only work for the first 16 colors.

As recommended in the linked bug, dimming should be done by overpainting with a mostly-transparent white or black (or whatever-the-background-color-is) in the ::draw handler, as done in the VTE test app.

mattrose:
So, I was playing around with this in my "minivte" playground that I use and got a semiworking POC, so I plugged it into terminator and came up with this.

It's terrible code, and needs cleanup, but it works.  Cleanup to come.

Fixes #74


mattrose:
This is a mess.  opening a new PR.

PR #196, opened 2020-08-26
Fix: Forbid Duplicate Key Bindings in `Preferences > Keybindings`

dkmvs:
This PR should fix #190.

**Changes:**
It's no longer possible to assign the same key binding to more than one action. For example, now when reproducing the steps in #190, the following error message will be displayed:
image

**Tests:**
I added `tests/test_prefseditor_keybindings.py` which tests that
1. Default key binding accelerators are distinct.
2. Message dialog appears when a duplicate key binding accelerator is attempted to be assigned to a different action.
3. Duplicate key binding accelerator cannot refer to more than one action.

One thing to keep in mind when running the tests locally is that the key bindings redefined in `~/.config/terminator/config` can conflict with the tests. That being said, the simplest way to avoid any possible conflicts is to remove  `~/.config/terminator/config`.

PR #224, opened 2020-10-04
Fix: Key Binding Clearing in `Preferences > Keybindings`

dkmvs:
This PR should fix incorrect clearing of key bindings in `Preferences > Keybindings`, which was introduced in #196.

**Steps to Reproduce:**
1. Open `Preferences > Keybindings`.
2. Clear any non-empty key binding, that is double click on any key binding which doesn't say `Disabled` and then press Backspace.
3. Try to assign a new key binding to any action - it won't work.

This happens because `Gtk.accelerator_parse`, which was used in #196, doesn't allow `None` to be used as a value. To fix this, the key binding value is now set to an empty string (`""`)  instead of `None` the moment it's been cleared.

I also added a test to reproduce the behaviour described in this PR.

### DIFF
--- a/terminatorlib/prefseditor.py
+++ b/terminatorlib/prefseditor.py
@@ -1745,7 +1745,7 @@ class PrefsEditor:
         liststore.set(celliter, 2, 0, 3, 0)
 
         binding = liststore.get_value(liststore.get_iter(path), 0)
-        self.config['keybindings'][binding] = None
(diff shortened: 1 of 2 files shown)


mattrose:
You are a pro.  I'll take a look at this this morning hopefully.  Thanks again

mattrose:
Awesome, thank you.

Issue #210, opened 2020-09-17
config settings lost when I cancel "Close multiple terminals" dialog


bitinerant:
I have a Terminator window that launches another Terminator window with a specific profile via `terminator -l ...`. This second window has 5 horizontal terminals. If I close the second window with the x in the top-right corner, I get the expected dialog "Close multiple terminals?". If I click 'Cancel' in this dialog, Terminator silently zeros out the `~/.config/terminator/config` file and displays this message in the first Terminator window:

```
ConfigBase::save: Unable to save config: 'ascii' codec can't encode character '\u21d2' in position 2252: ordinal not in range(128)
```

After this, opening a new Terminator window uses default settings. As far as I can tell, all config settings are permanently lost.

The U+21D2 mentioned in the error *is* displayed in several of the 5 horizontal terminals in the second Terminator.

I'm using Terminator 1.91 in Ubuntu 20.04.

mattrose:
can you attach a copy of your config?  It looks like it got corrupted somehow.

bitinerant:
Thanks for helping with this issue. :smiley: Unfortunately, I can't post the config publicly because it contains sensitive information. It may be helpful to know that `file` says it is `UTF-8 Unicode text`. The fist 3 bytes are 5b 67 6c, so no UTF marker. It *does* contain the U+21D2 character (hex bytes e2 87 92) in a few places, all of them on a `command = ...` line. Is this not supported?

mattrose:
Hmm, can you post the command that's causing it to fail?   Do you need those specific unicode characters in the command?  I just need some way of reproducing the bug to be able to debug it.

bitinerant:
I have found a way to reproduce what I think is the same bug, but without a customized config file:
* close all instances of Terminator
* delete or rename ~/.config/terminator/config
* open Terminator and select Preferences > Layouts > Terminal child1 and in the Custom command box type: `echo fiance; sleep 1h`
* click close and then close Terminator
* run Terminator again; it should display "fiance"
* repeat all of the above steps, but this time in the command box, copy-paste (note the accent on the 'e'): `echo fiancé; sleep 1h`
* I would expect it to act the same both times (except for the accented 'e'), but instead it opens with a default configuration and has a zero-byte ~/.config/terminator/config file

mattrose:
Thank you, that'll make it much easier to work on.

mattrose:
[...]

PR #235, opened 2020-10-08
Change how config file is saved

### LINKED ISSUE #234: symlinked config replaced with regular file
My terminator config file is symlinked to my dotfiles repo.
When I change a preference and close the window, the symlink is replaced with a regular file.

Introduced in
d10919df5fb4ff0b103c08c2e074f382982a7678

Also, this might affect file history/recovery, as a new inode is created.

planet36:
This PR changes how the config file is saved.

Current behavior:
1. write parser contents to config.tmp
2. rename config.tmp to config

Proposed behavior:
1. copy config to config~
2. write parser contents to config
3. remove config~ (this could be optional)

Fixes #234

### DIFF
--- a/terminatorlib/config.py
+++ b/terminatorlib/config.py
@@ -71,7 +71,6 @@ KeyError: 'ConfigBase::get_item: unknown key algo'
 """
 
 import os
-import pathlib
 import shutil
 from copy import copy
 from configobj import ConfigObj, flatten_errors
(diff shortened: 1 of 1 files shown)


mattrose:
Looks good, but 2 things:

1.  It doesn't create the file if it doesn't exist, so anyone who freshly installs terminator wouldn't be able to make config changes without touching the file first.

2. Any chance you could take a look at https://github.com/gnome-terminator/terminator/issues/210#issuecomment-696261273 and see if you can fix that issue, while you're looking at that section of the code?  It looks like, at the very least, we should catch the exception that was thrown, but if you think you can write the file properly that would be best.

If you can't fix the second issue that's fine, but I thought you might want to take a stab.   Let me know if you get stuck anywhere.  Thanks!

planet36:
Does Terminator have to work in Python 2.7?

mattrose:
Yes

mattrose:
Thanks for looking at #210, btw

mattrose:
Thanks for fixing this, I'll check it out either later today or tomorrow.

mattrose:
I like it.  Thanks so much for putting up with the requested changes.  It means a lot.

PR #249, opened 2020-10-17
Preserve searchbar case sensitivity state in config file on state change

yoavp77:
- Add a configuration option named 'case_sensitive' to determine whether the searchbar matches case by default
- Persists this state to the config file whenever it is changed so that it retained whenever terminator is relaunched

### DIFF
--- a/terminatorlib/config.py
+++ b/terminatorlib/config.py
@@ -122,6 +122,7 @@ DEFAULTS = {
             'putty_paste_style_source_clipboard': False,
             'smart_copy'            : True,
             'line_height'           : 1.0,
+            'case_sensitive'        : True,
         },
         'keybindings': {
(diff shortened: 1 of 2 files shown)


mattrose:
I love the idea.  I'll pull it down and test it very soon.

yoavp77:
I have another proposal to add an "Inverse Search" checkbox to the search dialog which will allow search from the bottom of the terminal history up, shall I introduce that in a separate PR? It also touches the same two files.

image

mattrose:
> I have another proposal to add an "Inverse Search" checkbox to the search dialog which will allow search from the bottom of the terminal history up, shall I introduce that in a separate PR? It also touches the same two files.
> 
> image

A separate PR for that would be great.

mattrose:
Looks great!  Thanks

PR #257, opened 2020-10-20
Add support for inverted search

yoavp77:
This change adds support for inverting the search direction - searching will optionally be from the bottom-up (meaning, recent terminal history will be matched first). 

Selected search direction is persisted in the config file so that it is retained the next time terminator is launched.


mattrose:
Thank you for this

PR #69, opened 2020-05-10
Update German and Croatian

milotype:
- both translations are complete

Issue #90, opened 2020-05-17
Introduce and document Transifex


lazyfrosch:
I created a Transifex organization, this should help with future translation efforts.

https://www.transifex.com/terminator

- [ ] Add documentation
- [ ] Add badge and info to README/Contributing
- [ ] Add links to about

comradekingu:
@lazyfrosch Transifex Doesn't help translation or libre software projects in the slightest.
It neither is libre software, nor does it support libre software. It functions as something you need libre software to combat. It isn't even a good tool…
Set up shop with Weblate instead. I'll help where I can :)

nerun:
Hi, could you add manpages to transifex? I would like to see terminator's linux page manual translated.

PR #287, opened 2020-11-14
Update German and Croatian

milotype:
Update, as discussed in issue **Update German and Croatian #69**, based upon current `terminator.pot` file.

BTW, for tranlsations, maybe you could consider using an online translation platform like Weblate, Poedit, Transifex, Launchpad (in the order of my preference). IMO, users are often potential translators! Sorry, I'm not a programmer, so I can't help with implementing it ...


mattrose:
@lazyfrosch is working on Transifex integration in #90, but I think he's fairly busy right now.  Thanks for putting up with us 😃

PR #295, opened 2020-11-19
Minor fixes

strottie:
Here are fixes for two minor fixes I stumbled on while doing some plugin development.
Cheers!


mattrose:
I actually exclaimed "OMG" when I saw the message on the 2nd commit as it is something that I've been meaning to fix basically since I got commit access, but was always too low on the priority list.   Thank you so much for this.

PR #309, opened 2020-12-01
Update README with install instructions

tomeksabala:
Added a hint on where to look for information on enabling terminator package repository.
Personally I couldn't find the new terminator in my Ubuntu 20.04 packages. Only the old terminator (1.9) was available.
There are instructions on how to add the custom ppa with binary for Ubuntu 20.04 in the  terminator/INSTALL.md.
Core README.md however makes it look like the INSTALL.md is only for source install.

It was just misleading, at least for me, and I was googling for a while on how to get this working having the information right under my nose.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ You can find the project on https://github.com/gnome-terminator/terminator
 
 ## Installing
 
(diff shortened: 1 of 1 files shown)


mattrose:
I like it, can you just make one small change?

Can you change "enable the package repository" to "enable an updated package repository"  I've read it over and I think it reads clearly.

Just to make clear that you're kind of straying off the straight and narrow path of the distribution by doing this

tomeksabala:
Sure. I've just done it. Thanks!

mattrose:
Thank you!!

PR #312, opened 2020-12-02
terminator/terminatorlib/config.py add shortcut key Alt-[1-0] to swit…

FineFan:
terminator/terminatorlib/config.py add shortcut key Alt-[1-0] to switch_to_tab_[1-10].


mattrose:
see #311

PR #325, opened 2020-12-21
fixed 'validate' module import for newer version of configobj

### LINKED ISSUE #324: ModuleNotFoundError: No module named 'validate'
Hi I'm getting this:

```
λ  ~  terminator                                                                                                                                                                                                                                                                                                                                                        1   10:56:06  
  File "/usr/sbin/terminator", line 48, in <module>
    import terminatorlib.optionparse
  File "/usr/lib/python3.9/site-packages/terminatorlib/optionparse.py", line 24, in <module>
    from . import config
  File "/usr/lib/python3.9/site-packages/terminatorlib/config.py", line 76, in <module>
    from validate import Validator
ModuleNotFoundError: No module named 'validate'
```

Here is my version of python:
```
Python 3.9.1
```
Also when I try to install the validate module:
```
 λ  ~  pip install validate                                                                                                                                                                                                                                                                                                                                                  10:59:39  
Defaulting to user installation because normal site-packages is not writeable
Collecting validate
  Using cached validate-1.0.1.tar.gz (32 kB)
    ERROR: Command errored out with exit status 1:
     command: /usr/bin/python -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/tmp/pip-install-0tvxkv2x/validate/setup.py'"'"'; __file__='"'"'/tmp/pip-install-0tvxkv2x/validate/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base /tmp/pip-pip-egg-info-6pyjqa9s
         cwd: /tmp/pip-install-0tvxkv2x/validate/
    Complete output (8 lines):
      File "<string>", line 1, in <module>
      File "/tmp/pip-install-0tvxkv2x/validate/setup.py", line 13, in <module>
        from configobj import __version__ as VERSION
      File "/tmp/pip-install-0tvxkv2x/validate/configobj.py", line 1632
        except Exception, e:
                        ^
    SyntaxError: invalid syntax
    ----------------------------------------
ERROR: Command errored out with exit status 1: python setup.py egg_info Check the logs for full command output.
```
It looks like it's Python 3.9.1 not compatible?

KenN7:
In the last version of python-configobj (5.0.6-10), the path to the validate module has changed, this fixed #324.
For now downgrading to 5.0.6-9 fixes the problem, but it probably needs to be updated in the future

### DIFF
--- a/terminatorlib/config.py
+++ b/terminatorlib/config.py
@@ -74,7 +74,10 @@ import os
 import shutil
 from copy import copy
 from configobj import ConfigObj, flatten_errors
-from validate import Validator
+try:
+    from validate import Validator
+except ModuleNotFoundError:
+    from configobj.validate import Validator
 from .borg import Borg
 from .util import dbg, err, DEBUG, get_system_config_dir, get_config_dir, dict_diff
 


mattrose:
The test fails, probably because the configobj module on the test server doesn't have the new version yet.  I have no idea how to fix this, or whether I should.

KenN7:
Yes, it's probably the test server not up to date yet, i'd say to wait until major distros and package manager have the latest version.
Then, if it you feel its ok to leave this PR open, it could be merged when the tests pass eventually.

mattrose:
Yeah, I'll wait for a bit and see if the test environment updates.  I may also consult with the configobj developers to see if there's a better way to do this test.

Thank you so much for submitting this PR, sorry to be so short in my first response, I really do appreciate it so much.

planet36:
Would this suffice?
```python
try:
    from validate import Validator
except ModuleNotFoundError:
    from configobj.validate import Validator
```

mattrose:
That looks good.  @KenN7 can you make the change? Then, assuming that passes, I can pull that in.

KenN7:
Good one, thx @planet36

C-mmon:
Yes, that works for me.
Thank you very much @planet36

chindit:
Tested and working.

KenN7:
any updates on this ? i have no idea why the tests timeout ?

mattrose:
The tests timed out because of some other code that was in master.  I actually pulled that code out and tests are running again.  Can you update this branch, and then the tests should run.  Also, I believe this happens only on Arch, and it has to do with something in the arch configobj code, so I'm a little hesitant to fix it in Terminator only to have it break on other platforms.

KenN7:
@mattrose I already updated the code with the fix of @planet36, (the try.. except..), is there something else to do ?
True, the problem is now only on arch, but the current fix (with try, except) doesnt break other distros.

mattrose:
So, looking at the arch package, between 5.0.6-9 and 5.0.6-10 they went from using a tarball to using a git checkout, and then 3 days later made another change and released [5.0.6.r110.g3e2f4cc-1](https://github.com/archlinux/svntogit-community/commit/6796dda1d14ac19d71b2ae7d06d8dfa39388a237#diff-3e341d2d9c67be01819b25b25d5e53ea3cdf3a38d28846cda85a195eb9b7203a) which looks like it [fixes the original problem](https://github.com/archlinux/svntogit-community/blob/6796dda1d14ac19d71b2ae7d06d8dfa39388a237/trunk/PKGBUILD#L34)  at the source

Sorry to be such a pain about this, but I'd rather not have Terminator work around an arch-specific problem that is now fixed, and have a weird try...except clause live on in the code that in 6 months nobody will understand why it's there.

Can you try upgrading to  python-configobj 5.0.6.r110.g3e2f4cc-1 and see if that fixes the problem, rather than the patch?

KenN7:
@mattrose allright, just tried with  5.0.6.r110.g3e2f4cc-1 and it does fix the problem, so i guess this can be closed for now :)
Cheers.

mattrose:
Thanks for looking into this and coming up with a solution, sorry we couldn't incorporate it.

PR #333, opened 2020-12-27
Feature: Relaunch command option on held open after child exit

zothar:
This is a forward port of a feature I added to a local 0.97 copy a number of years ago after switching away from Gnome Terminal.  This approach was what I figured out first, so it may not be the best way to implement the feature (the relaunch option in the right-click menu, for example), so I'm just putting this out there and wouldn't be surprised if it needs further development before merging.

I don't recall for sure, but Terminal.relaunch_command might have been needed for some use case I've now forgotten the details to.   I've not tried re-testing without it.

### DIFF
--- a/terminatorlib/terminal.py
+++ b/terminatorlib/terminal.py
@@ -126,7 +126,10 @@ class Terminal(Gtk.VBox):
     custom_encoding = None
     custom_font_size = None
     layout_command = None
+    relaunch_command = None
     directory = None
+
+    is_held_open = False
 
     fgcolor_active = None
     bgcolor = None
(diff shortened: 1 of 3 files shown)


mattrose:
Thank you so much for this.  This actually is a really good way of fixing an issue someone raised.  How would you set `config['exit_action']` just through manually editing the config file?  If so, would you feel comfortable adding a UI in the preferences general tab?

zothar:
If you're talking about adding UI for managing config['exit_action'], it's already there as part of the Command tab of the Profile as "When command exits:".

If you're talking about adding UI for something else in the preferences, please clarify.  For now, I'll assume I can sort it out once I've got a better idea of what you're asking.

mattrose:
Whoops, missed that when I was looking through the preferences dialog!  Thanks, looks great.  Merging now

Issue #323, opened 2020-12-16
"Copy email address" actually doesn't quite do that


axelsimon:
Hi,
This is a simple "papercut" type issue: when right-clicking an email address in Terminator, there's a "Copy email address" option.
Except that instead of doing just that, it prepends it with `mailto:`.
So if you try and copy axel@example.com, your buffer actually gets filled with `mailto:axel@example.com` and not `axel@example.com`.

A small thing for sure, but i find the menu entry not quite right, so i wanted to report this.

Thanks!

PR #335, opened 2020-12-27
Fixing #323

JAugusto42:
I'm not sure if this is the best way to do this, but I don't see any other use for this condition.


mattrose:
gonna close this one because I'm assuming it's superseded by #338

PR #339, opened 2020-12-27
remove rewrap on resize from option and remove functions too

JAugusto42:
@mattrose What do you think?


mattrose:
Looks fantastic!  Thank you

Issue #216, opened 2020-09-23
Use auto-hiding scrollbar from GTK3, rather than the old one


mattrose:
From a comment on reddit

Handling of scrollbar - Tilix has perfect solution (auto-hiding, barely visible scrollbar - using modern Gnome style); Terminator has old scrollbar from GTK2 times, with an option to disable it completely (but no option for autohide the same way Tilix does it).

PR #340, opened 2020-12-28
feat: Migrating from hbox to GtkBox

JAugusto42:
I try to solve #216, but it was more difficult than I imagined. So I just updated the code.


mattrose:
This is a great update, thanks

Issue #244, opened 2020-10-13
FIXED: Install terminator from repositories


dragonxtek:
When I try to install it from repositories, the following issues appear:

error: can't copy 'data/terminator.desktop': doesn't exist or not a regular file
error: can't copy 'data/terminator.metainfo.xml': doesn't exist or not a regular file

To solve it I need to do the following:

cd data/terminator.desktop.in data/terminator.desktop
cp data/terminator.appdata.xml.in data/terminator.metainfo.xml

mattrose:
What repositories are you trying to install from?

dragonxtek:
I did the following:
git clone https://github.com/gnome-terminator/terminator.git
cd terminator
sudo python3 setup.py install

mattrose:
You need to run `python setup.py build` first.  
Also, you should run `python3 setup.py install --single-version-externally-managed --record=install-files.txt` otherwise a bunch of assets won't get copied properly.

dragonxtek:
I re-installed as you said, but now I can't run terminator:

```
...
writing terminator.egg-info/PKG-INFO
writing dependency_links to terminator.egg-info/dependency_links.txt
writing requirements to terminator.egg-info/requires.txt
writing top-level names to terminator.egg-info/top_level.txt
reading manifest file 'terminator.egg-info/SOURCES.txt'
reading manifest template 'MANIFEST.in'
warning: no previously-included files found matching 'po/.intltool-merge-cache'
writing manifest file 'terminator.egg-info/SOURCES.txt'
Copying terminator.egg-info to /usr/local/lib/python3.6/dist-packages/terminator-2.0.1-py3.6.egg-info
running install_scripts
changing mode of /usr/local/bin/terminator to 755
changing mode of /usr/local/bin/remotinator to 755
writing list of installed files to 'install-files.txt'
> terminator              
Traceback (most recent call last):
  File "/usr/local/bin/terminator", line 4, in <module>
    __import__('pkg_resources').run_script('terminator==2.0.1', 'terminator')
  File "/home/dragonxtek/.local/lib/python3.6/site-packages/pkg_resources/__init__.py", line 666, in run_script
    self.require(requires)[0].run_script(script_name, ns)
  File "/home/dragonxtek/.local/lib/python3.6/site-packages/pkg_resources/__init__.py", line 1454, in run_script
    .format(**locals()),
[...]

PR #342, opened 2020-12-28
Local user install: terminator.desktop and terminator.metainfo.xml

zothar:
Partial #244 fix


zothar:
The inclusion of 4a3dc0c is apparently due to my newb-ness.

mattrose:
That looks good, merging in

PR #349, opened 2020-12-31
Fix #345 The Alt+L layout chooser is too small

### LINKED ISSUE #345: The Alt+L layout chooser is too small
When I press `Alt+L` on terminator, I get this window:
image

But it is too small and I can barely see my layouts.

I tried resizing it, but the new size is forgotten on the next time I open it.

**Desktop (terminator 1.91)**
 - OS or Linux Distribution: Ubuntu
 - Version 20.04.1

evandrocoan:
I am not sure, but directly setting the size on Python fixed the issue in my platform.

I am also increasing the height +50 pixels because a rectangular size seems more fit than a square box.

image


mattrose:
Looks good to me.

PR #350, opened 2020-12-31
Update terminator.appdata.xml.in

jooola:
This commit is based on a patch from Debian Salsa repository. See https://salsa.debian.org/python-team/packages/terminator/-/blob/7c08e04fb205a61e41110b3f274298c9c660efd1/debian/patches/appdata_improvement.patch

- Changed screenshots urls from http to https
- Updated the license identifier to match [SPDX licence list](https://spdx.org/licenses/) (the one on the debian patch is a deprecated identifier)

### DIFF
--- a/data/terminator.appdata.xml.in
+++ b/data/terminator.appdata.xml.in
@@ -3,7 +3,7 @@
 <component type="desktop">
  <id>terminator.desktop</id>
  <metadata_license>CC0-1.0</metadata_license>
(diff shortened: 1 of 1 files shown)


dmaphy (approved):

lazyfrosch (commented):
Thanks for adopting my patch :wink: 

I guess we should move the images to gnome-terminator.org at some point.

But thanks for now.

PR #351, opened 2021-01-02
German language updated

Brambleberry4:
Hopefully improved the German po-file.


po/de.po
```diff
@@ -185,24 +186,24 @@ msgid ""
...
 
 #: ../terminatorlib/container.py:198
 msgid "Do not show this message next time"
-msgstr "Diese Nachricht das nächste Mal nicht anzeigen"
+msgstr "Diese Nachricht in Zukunft nicht mehr anzeigen"
 
 #: ../terminatorlib/encoding.py:34
 msgid "Current Locale"
-msgstr "Derzeitige Standorteinstellungen"
+msgstr "Derzeitiger Standort"
```
dmaphy, line 206: Probably it's more a "Derzeitige Spracheinstellungen"?

po/de.po
```diff
@@ -706,11 +707,11 @@ msgstr "Senkrechter Strich"
 
 #: ../terminatorlib/preferences.glade.h:26
 msgid "GNOME Default"
-msgstr "GNOME-Vorgabe"
+msgstr "GNOME-Standard"
 
 #: ../terminatorlib/preferences.glade.h:27
 msgid "Click to focus"
-msgstr "Klicken zum Fokussieren"
+msgstr "Klicken zum fokussieren"
```
dmaphy, line 714: I'd leave the big F here.

po/de.po
```diff
@@ -914,19 +915,19 @@ msgstr "Titelleiste unten (Neustart erfordelich)"
 
 #: ../terminatorlib/preferences.glade.h:79
 msgid "Hide size from title"
-msgstr "Größe im Titel verstecken"
+msgstr "Größe vom Titel verstecken"
```
dmaphy, line 918: I think "im" is more correct here.

po/de.po
```diff
@@ -1014,7 +1015,7 @@ msgstr "Allgemein"
 
 #: ../terminatorlib/preferences.glade.h:104
 msgid "_Run command as a login shell"
-msgstr "Befehl wie eine Login-Shell _ausführen"
+msgstr "_Befehl als Login-Shell ausführen"
```
dmaphy, line 1018:
```suggestion
msgstr "_Befehl in Login-Shell ausführen"
```

po/de.po
```diff
@@ -103,11 +104,11 @@ msgid ""
 "terminals in grids (tabs is the most common default method, which Terminator "
 "also supports)."
 msgstr ""
-"Werkzeug für erfahrene Nutzer, um Terminals anzuordnen. Es ist inspiriert "
-"von Anwendungen wie: gnome-multi-term, quadkonsole, usw., deren Hauptfokus "
-"auf dem Anordnen der Terminals in einem Gitter liegt (die Verwendung  von "
-"Reitern ist die meist verbreitete Methode, welche von Terminator auch "
-"unterstützt wird)."
+"Ein Werkzeug für erfahrene Nutzer, um Terminals anzuordnen. Es ist von "
+"Anwendungen wie: gnome-multi-term, quadkonsole, usw. inspiriert, deren "
```
dmaphy, line 108:
```suggestion
"Anwendungen wie gnome-multi-term, quadkonsole, usw. inspiriert, deren "
```

po/de.po
Brambleberry4: I am not sure about the correct translation of "Locale"

Brambleberry4: Do you know about the context?

dmaphy: Usually the context for the use of the term "locale" is about language or charset encoding. That's why I was suggesting "Spracheinstellungen". An alternative which may be even better would be "(momentaner) Zeichensatz" or something similar.

A quick look into the code told me that this translation is about the context menu when you right click somewhere in the terminal, there is a menu point "Encodings" which is most properly translated as "Zeichensatz" in my opinion.

Brambleberry4: so I found https://de.wikipedia.org/wiki/Locale where Locale is the umbrella term for all these settings location language, keyboard layout etc.. They translate it with "Einstellungssatz".

dmaphy: "Einstellungssatz"… yeah, I like that word, even if it sounds a bit unusual to me personally. But seems it's the most proper translation then. (:

Brambleberry4: Okay I will use that :)

lazyfrosch (approved):
LGTM

lazyfrosch:
Thanks for your work, I will document the use of Transifex soon, so it will be easier to help translating without the need to work on PO files.

PR #353, opened 2021-01-03
Update RELEASE docs

### LINKED ISSUE #348: Rethinking release artifacts
While updating terminator in Debian I wasn't really to use the sdist tarball, as it is missing multiple files, including:

* po files
* icons
* stylesheets
* some tests
* license and other metadata

We should probably do a git archive on release and sign this.

Anyways, great work on the release @mattrose :heart: 

Current doc:
https://github.com/gnome-terminator/terminator/blob/v2.0.1/RELEASE.md#signed-artifacts

Diff:
https://salsa.debian.org/python-team/packages/terminator/-/commit/45f77a074ea1f0e4919a01047c434ddf6354c5e5

lazyfrosch:
fixes #348

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -63,8 +63,14 @@ git push --tags
 To provide a signed tarball for distributions we use sdist and gpg:
 
 ```
-python setup.py sdist
-gpg -u markus@lazyfrosch.de --armor \
+VERSION=$(git describe --tags | sed s/^v//)
+GPGKEY=$(git config --get user.email)
+
+mkdir -p dist
+
+git archive HEAD --prefix terminator-${VERSION}/ -o dist/terminator-${VERSION}.tar.gz
+
+gpg -u ${GPGKEY} --armor \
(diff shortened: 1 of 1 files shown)


RELEASE.md
```diff
@@ -63,8 +63,14 @@ git push --tags
 To provide a signed tarball for distributions we use sdist and gpg:
 
 ```
-python setup.py sdist
-gpg -u markus@lazyfrosch.de --armor \
+VERSION=$(git describe --tags | sed s/^v//)
+GPGKEY=markus@lazyfrosch.de
```
mattrose, line 67: would something like `GPGKEY=$(git config --get user.email)` work for you?  If you're in the terminator checked out directory it should pull your correct email.  That way we can both just copy and paste from the document.

lazyfrosch: Nice idea, done

mattrose (approved):

Issue #79, opened 2020-05-14
add or merge Tmux control mode support 


yosefrow:
Tmux control mode would allow users to connect terminator to a tmux session.

As I understand this feature has already implemented at https://github.com/manno/terminator .

I added an issue to that repo as well. So maybe the author will respond and look into merging his code into this main repo. 

But if for some reason he is not able to do so, I want to at least raise awareness that the feature seems to exist for those who are interested.

mattrose:
This looks really good.  Hopefully the user who owns the fork will craft a proper pull request against this repo.  If not, we'll still look at it, but I personally don't use tmux that much, so any guidance you could give would help.  What are people looking for from tmux integration?

mattrose:
Looking further.  Wow, that's a lot of work they've done.  Hopefully they'd be willing to work with us, as that terminator code is probably missing all the python3 fixes.

yosefrow:
@mattrose 

TLDR:

I'm sure there are more features that a more advanced user might know about, for me I'm looking for the following:

- full integration with tmux windows/panes representation (resizing, display, mouse integration)
- integration with tmux scroll buffer and os clipboard
- integration with tmux pane/operations (e.g. rotate panes, maximize pane etc)
- ability to connect to local or remote tmux session
- integration with tmux status bar (nice to have!)

As an intermediate level tmux user , I think I can best provide an insight by describing my use of tmux.

Because I administrate multiple systems remotely, I find myself more often than not logging into the remote server and then running tmux inside terminator, because otherwise I'd have to ssh many times (every time I opened a tab). Opening tmux inside the remote server, saves me time and headache associated with constantly connecting. But it also allows me to save my sessions remotely and reconnect to them the next day, or between laptop reboots.

Sometimes, even though I have no work to do on the remote server, I actually open a tmux session on the remote server, just because I want to continue where I left off the next day. I know that there are some plugins that allow you to save terminator sessions. But so far I haven't found one that accurately saves the buffer of all active windows/tabs and the window/tab configurations.

Additionally, I appreciate the fact that the tmux status bar allows you to dynamically and contextually display any text you want. It is the closest I've seen to a "task bar" that you typically find on desktop systems and it adds just enough extra benefit, that I can permanently live in my terminal in full screen. For example, you can place small scripts that will add a kubernetes or git context indicator for your current folder, hostname, cpu info, memory usage, clock etc.

[...]

PR #378, opened 2021-01-31
Tmux Control Mode (squashed)

manno:
This is the tmux code from 2016 applied on top of the current master branch.

It's not functional yet.

(for https://github.com/gnome-terminator/terminator/issues/79)

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -8,7 +8,7 @@ Packages are known to be available under the name "terminator" under a
 lot of distributions, see below for a list.
 
 I also maintain a PPA for Ubuntu 20.04 and up that has the latest release
-If you're running ubuntu 20.04 or later, you can run 
+If you're running ubuntu 20.04 or later, you can run
 
 ```
 sudo add-apt-repository ppa:mattrose/terminator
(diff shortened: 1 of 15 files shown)


mattrose:
Wow, thanks for this.  I'll pull this in and take a look at it as soon as I can, and hopefully I can help you work through the remaining issues.

mattrose:
Seems to work so far, and the test failing is a simple fix.  What specifically is not working?

manno:
> Seems to work so far, and the test failing is a simple fix. What specifically is not working?

So back then, the decision was made to develop tmux support mostly independent of the main code path.

Clients like iterm2 detect tmux control sequences and open a new window (`echo -e '\x1bP1000p%begin 1612615231 294 0\n%end 1612615231 294 0\n'`), but terminator would wrap all of that.
I wish it was different, as this means dealing with [SSH and tmux binaries](https://github.com/gnome-terminator/terminator/pull/378/files#diff-80ad70535f58a538410efcf1db8fc2324ad9369634cad152f6bd4d9fa4e9db06R13).

It also means, the code in this PR won't execute unless terminator is started with `./terminator -t`. And that code path currently runs in to an error:

```
  File "/home/mm/co/terminator/terminatorlib/terminal.py", line 1326, in do_deferred_on_vte_size_allocate
    self.on_vte_size_allocate(widget, allocation)
  File "/home/mm/co/terminator/terminatorlib/terminal.py", line 1329, in on_vte_size_allocate
    column_count = self.vte.get_column_count()
AttributeError: 'NoneType' object has no attribute 'get_column_count'
```

I'm not sure how to fix that, but if it's complicated to fix, I'd suggest to just detect the escape sequences and react to them and throw all the popen ssh/tmux code away.

mattrose:
Just checking in to say I'm pretty busy starting a new job right now, but I'll take a look as soon as I can.  I haven't forgotten about this.

iamjackg:
I actually think the "bypass" approach makes a lot of sense in terms of usability since it would otherwise require an inter-window communication method and add more UX concerns. I'm hoping to find some time this weekend to go through the MR and get it working :)

iamjackg:
So I've spent a couple of days getting this working. It's _mostly_ usable at a basic level, but only if there is only one client connected at any one time. In short, the code is able to push all (most) state changes from Terminator to tmux, but it's not able to _react_ to changes made by either an attached tmux instance or other Terminator instances attached to the same session.

If your workflow only involves having one backing headless instance of tmux running at all times and connecting/disconnecting one Terminator instance at a time, then this should be fine.

There are a few really gnarly incompatibilities between Terminator's features and what tmux allows. As an example, in Terminator you can change the zoom level of all panes independently, which is impossible in tmux. Doing it leads to very odd behaviour where a terminal appears to have a certain height×width, but the backing tmux pane is actually smaller/bigger, and text wraps in odd ways in the other panes.

I'll do some more cleanup and submit a separate PR. Maybe we can still include it with some caveats. I wish I had a Mac to test how iTerm2 deals with all of this.

mattrose:
I'm going to close this one, as we're now working on #408

PR #388, opened 2021-02-19
Fixed race condition when calling grab_focus after underlying vte could be closed

TheBigS:
I found a race condition where the command I specified for the new vte had already exited by the time `grab_focus` was called. This caused the terminal to start cleaning up and the `self.vte` was `None` by the time `grab_focus` executed.

This caused a `Could not call method 'has_focus()' on NoneType` error. Not a huge deal because the terminal closed itself anyway. Easy fix to avoid the error by adding the None check before the `has_focus` call.


mattrose:
Thanks for this

PR #389, opened 2021-02-19
Added new get_focused_terminal dbus command which returns uuid of current focused terminal

TheBigS:
I found `has_focus` wasn't working but `self.terminator.last_focused_term.uuid.urn` seems to be working. I also updated remotinator to support new `get_focused_terminal` ipc command.


mattrose:
I'll try to take a look at this this weekend

mattrose:
Sorry for making you wait so long.  This looks great, thanks!

PR #398, opened 2021-03-04
add feat: config to open links with single click

nojhan:
Instead of Ctrl-left click.

Note: I only have glade 3.22.2 instead of the 3.36.0 used for the previous setup.
I guess that's why there have been some editions in the XML, apart for the new checkbox.
I have absolutely no idea if that's a problem or not… the window works well on my Ubuntu 20.04, at least.

### DIFF
--- a/terminatorlib/config.py
+++ b/terminatorlib/config.py
@@ -125,6 +125,7 @@ DEFAULTS = {
             'line_height'           : 1.0,
             'case_sensitive'        : True,
             'invert_search'         : False,
+            'link_single_click'     : False,
         },
         'keybindings': {
             'zoom_in'          : '<Control>plus',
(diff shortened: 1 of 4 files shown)


mattrose:
This is definitely a feature I've wanted to add for a while!  Thank you.  I've been really busy but this is at the top of my to-merge list

mattrose:
I get this error when trying to check the box

  File "/Users/mattrose/Code/terminator/terminatorlib/prefseditor.py", line 1426, in on_link_single_click_toggled
    self.config['link_single_click'] = widget.get_active()
NameError: name 'widget' is not defined

mattrose:
And it doesn't seem to set the config variable properly.  Can you take a look at this again?

mattrose:
sorry, clicked wrong button there.

mattrose:
Looks good, merging

PR #399, opened 2021-03-04
Feat+run cmd on match

nojhan:
Adds the necessary features to add a regexp/command plugin.
My use-case is to be able to open a file at a specific line in my IDE, just by clicking on an error message.

For instance, for opening a python script indicated in a python error, within vim, use the following configuration with the RunCmdOnMatch plugin implemented in this PR:
- regexp = `\B(/\S+?\.py)\S{2}\sline\s(\d+)`
- command = `gvim --servername IDE --remote +{1} {0}`

This adds three interconnected features: 
- a plugin which can run any command when clicking on a link inserted on texts matching a regexp,
- a way to hide plugins in the preference/plugins window (by prefixing their name with an underscore),
- a way to bypass the URL handler (xdg-open), when the URLhandler plugin returns a `terminator://` URL.

The overall architecture is not really elegant, but it is the smallest footprint I found for implementing a match/command plugin.

It needs a URLhandler metaclass and will instantiate on-the-fly one plugin per configured match/command pair (hence the necessity to hide them, as the user is not supposed to interact with them anyway).
As those hidden plugins are responsible for "opening" the URL they match, a way to bypass the URL handler was needed. Returning `None` already had a semantic, so I opted for a fake protocol in the URI.

The GUI code is almost a copy/paste of the custom commands plugin, so it would be better to have the plugin window's entry point (instead of the contextual menu) and keep a single code base.

I'm not sure where to edit a user documentation.

### DIFF
--- a/terminatorlib/plugins/run_cmd_on_match.py
+++ b/terminatorlib/plugins/run_cmd_on_match.py
@@ -1,0 +1,522 @@
+import re
+import os
+import sys
+import subprocess
+
+from gi.repository import Gtk
+from gi.repository import GObject
+
+from terminatorlib.util import dbg
+import terminatorlib.plugin as plugin
+from terminatorlib.config import Config
+from terminatorlib.translation import _
+from terminatorlib.util import get_config_dir, err, dbg, gerr
+
+(CC_COL_ENABLED, CC_COL_REGEXP, CC_COL_COMMAND) = list(range(0,3))
+
+AVAILABLE = ['RunCmdOnMatchMenu']
+
+# For example, open scripts names outputed by python in Vim at the given line number:
+# match = r'\B(/\S+?\.py)\S{2}\sline\s(\d+)' # Python's log file matching
+# cmd = "gvim --servername IDE --remote +{1} {0}"
+
+# This class is not useful as is, it needs to be forged through MetaRCOM metaclass to be useful
+# (because the API use static class properties).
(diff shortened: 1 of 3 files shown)


mattrose:
Sorry for letting this drop for so long.  I actually love this feature and I think the code changes are mostly good.  

I checked it out and tried to use the example you put in the description, but I can't seem to figure out how to click on the text so that it brings up the command.  It should underline the line and that should let me click on it, but it doesn't seem to do that.  Is there something I'm missing?

nojhan:
The full test procedure would goes:
Add the `\B(/\S+?\.py)\S{2}\sline\s(\d+)` regexp, linking to your favorite editor, `gvim +{1} {0`.

Run a script that display the info to catch, here a Python script displaying THE FULL PATH.
```sh
echo "print('What the F)" >  test.py
python3 $(pwd)/test.py
```
Note that if you just do `./test.py`, then the path information is contextual and it's not going to work (or else we would need to pass the current directory from the shell to the URL manager).

Then, you have to pass the mouse over the path segment and [Ctrl-] click.

mattrose:
Huh. that's what I thought I did, but I'll try again.

Tried again.  Worked.  It may have been my gvim command.  Merging.  Thanks so much for this, and sorry it took me so long to get around to looking at it.

nojhan:
Awesome, thanks!
-- 
nojhan

PR #406, opened 2021-03-21
i18n: pt_BR: add missing space in translations with shortcuts

terceiro:
I know translations are supposed to be on transifex but this is just a drive-by fix for issues I found when using terminator, I don't plan to work on translations in the long term.


mattrose:
Thank you!

PR #408, opened 2021-03-21
Tmux control mode v2

iamjackg:
This is a cleanup of #378. A lot of changes went into this.

I spent the last couple of days getting it working. It's mostly usable at a basic level, but only if there is only one client connected at any one time. In short, the code is able to push all (most) state changes from Terminator to tmux, but it's not able to react to changes made by either an attached tmux instance or other Terminator instances attached to the same session.

If your workflow only involves having one backing headless instance of tmux running at all times and connecting/disconnecting one Terminator instance at a time, then this should be fine.

There are a few really gnarly incompatibilities between Terminator's features and what tmux allows. As an example, in Terminator you can change the zoom level of all panes independently, which is impossible in tmux. Doing it leads to very odd behaviour where a terminal appears to have a certain `height`x`width`, but the backing tmux pane is actually smaller/bigger, and text wraps in odd ways in the other panes. I ended up disabling zooming and drag&drop altogether when in tmux mode.

There are other issues: tmux's window size includes all the pane separators, so if you set a client size of 80x20 and then split it in two you don't get two 40x20 terminals, but one 39x20 and one 40x20, because the vertical separator takes up a column. Terminator determines its window size **excluding** the separator(s). I tried to add functionality to keep them in sync, but this doesn't always work.

Feel free to give it a try and see if it works fine for you.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,5 @@ terminatorlib/meliae
 .intltool*
 data/terminator.appdata.xml
 data/terminator.desktop
+
+venv/
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -8,7 +8,7 @@ Packages are known to be available under the name "terminator" under a
 lot of distributions, see below for a list.
 
 I also maintain a PPA for Ubuntu 20.04 and up that has the latest release
-If you're running ubuntu 20.04 or later, you can run 
+If you're running ubuntu 20.04 or later, you can run
 
 ```
 sudo add-apt-repository ppa:mattrose/terminator
@@ -24,13 +24,19 @@ dependencies yourself:
 **Python 3.5+ recommended:** `python3` or `python37` (in FreeBSD)
 
 **Python GTK and VTE bindings:**
-     
-     Fedora/CentOS: python3-gobject python3-configobj python3-psutil vte291 
+
+     Fedora/CentOS: python3-gobject python3-configobj python3-psutil vte291
                     keybinder3 intltool gettext
-     Debian/Ubuntu: python3-gi python3-gi-cairo python3-psutil python3-configobj 
-                    gir1.2-keybinder-3.0 gir1.2-vte-2.91 gettext intltool dbus-x11 
-     FreeBSD:       py37-psutil py37-configobj keybinder-gtk3 py37-gobject3 gettext 
+     Debian/Ubuntu: python3-gi python3-gi-cairo python3-psutil python3-configobj
+                    gir1.2-keybinder-3.0 gir1.2-vte-2.91 gettext intltool dbus-x11
+     FreeBSD:       py37-psutil py37-configobj keybinder-gtk3 py37-gobject3 gettext
                     intltool libnotify vte3
+
+**Python PyParsing library (only required for Tmux mode):**
+
+     Debian/Ubuntu: python-pyparsing
+
+     FreeBSD: devel/py-pyparsing
 
 If you don't care about native language support or icons, Terminator
 should run just fine directly from this directory, just:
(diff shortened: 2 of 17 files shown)


iamjackg:
Ah, some tests are failing :laughing: Didn't even notice those. I'll take a look later!

mattrose:
This is huge, but I think this'll be a real differentiator.  I don't personally use tmux, so I may ask a bunch of questions in the process of checking it out.  I have a mac with iTerm2 as well, so I may be able to help you out there.  I'm usually hanging out on https://gitter.im/gnome-terminator/community if you want to chat about this.  It's going to take a while to go through all the code, and just make sure I can understand at least the interface to the rest of terminator, if not understand the tmux code.

Thank you so much for this, it's why I love working on open source projects.

iamjackg:
No worries! Honestly the majority of the credit goes to Manno and the original contributors. 

I still feel like this approach might not be right and we might want to switch to a model where Tmux is driving the entirety of the logic, but that's a bit too far for me and my limited free time! :P

mattrose:
Maybe you, or somebody else will come along and pick it up later and make more improvements, but If we can get this to a fit state where it can be released, that's a huge step, and will make it easier for later contributors.  Thanks again.

mattrose:
Yay, tests pass!

iamjackg:
Hey @mattrose! Did you get a chance to try this out?

mattrose:
Sorry I've been busy lately, and I wiped my normal terminator development machine, and it dropped off my radar a bit.  I'll see if I can take a look today or tomorrow.

mattrose:
hmm, I get this error when running terminator -t 

  File "/home/mattrose/Code/terminator/terminator", line 50, in <module>
    from terminatorlib.terminator import Terminator
  File "/home/mattrose/Code/terminator/terminatorlib/terminator.py", line 19, in <module>
    from .tmux import notifications
  File "/home/mattrose/Code/terminator/terminatorlib/tmux/notifications.py", line 4, in <module>
    from terminatorlib.tmux import layout
  File "/home/mattrose/Code/terminator/terminatorlib/tmux/layout.py", line 1, in <module>
    from pyparsing import *
ModuleNotFoundError: No module named 'pyparsing'

Is that a dependency for tmux?

mattrose:
Installed pyparsing and noticed one glaring thing which is that debug output is outputted constantly.
 
 Also, I'm not sure how to use it with tmux. I tried `terminator -t` but that ... didn't work.  It opened terminator, but it left the terminal in a weird state
 
 I tried running tmux in a separate window, and tried terminator -t, and that just hung.

iamjackg:
> Is that a dependency for tmux?

Ah yes, my bad -- I'll add it to `setup.py`.

As for the rest I'll try to take a look today. I started it again today and it's just... not working. Sigh. I might end up re-implementing this from scratch. And sorry about the debug messages, I thought I had removed all the useless ones during a cleanup, but I missed a bunch.

ya-isakov:
@iamjackg Thank you a lot for working on this feature, which could be killer-feature for terminator. I've tried this MR today, and yes, terminator runs some tmux'es, but I cannot type anything. When I open new tab, I can see it in `tmux a`, and when I type something in `tmux a`, I can see it in terminator. Terminator is not opening new tab, when I crate one in `tmux a`.

iamjackg:
Just an update -- I haven't forgotten about this, I just haven't had much time to work on it. I started a complete rewrite that separates out the tmux control code into a separate library, because in its current form it's way too tied to the rest of the terminator code.

mattrose:
Thanks so much for staying with this.

iamjackg:
I left a comment about this on #79.

PR #416, opened 2021-04-08
Fixed geometry hints

Vulcalien:
I tried to use Window geometry hints, but it didn't work.
I tested this on: Debian 11 (all updates), Debian 10 and Ubuntu 20.04.

Details:
it seems that "set_geometry_hints" wants "None" as first argument.
Also, it's necessary to disable geometry hints, if we want to avoid restarting terminator.
So I wrote a little function that does that.

### DIFF
--- a/terminatorlib/terminal.py
+++ b/terminatorlib/terminal.py
@@ -1330,6 +1330,9 @@ class Terminal(Gtk.VBox):
         if self.config['geometry_hinting']:
             window = self.get_toplevel()
             window.deferred_set_rough_geometry_hints()
+        else:
+            window = self.get_toplevel()
+            window.disable_geometry_hints()
 
     def on_vte_notify_enter(self, term, event):
         """Handle the mouse entering this terminal"""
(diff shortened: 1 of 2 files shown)


mattrose:
Can you give me an idea on how to test this?

Vulcalien:
Sure!
So to reproduce the bug:
- I go to Terminator Preferences, select "Window geometry hints" and close Preferences
- try to resize the Terminator window

On all the distros I listed, it just ignores the geometry hints and resizes "smoothly".
Even if I restart Terminator, still same problem.

The problem should not be the config file, because it contains this:
```
[global_config]
  geometry_hinting = True
  ...
```

With my modifications:
- all Terminator instances must be closed (I guess it runs the older version otherwise)
- again, select "Window geometry hints" in Preferences
- I do `python3 terminator` to run it

It should work and resize correctly.

Vulcalien:
Also, some additional details: I tested the new version with the same config file and also after deleting the config file. It works on both cases.

About the disable_geometry_hints function:
without calling that function, if I go to Preferences and uncheck "Window geometry hints" I would need to restart Terminator to restore the "smooth" resize.

About using `None` instead of `self`:
I've read that around GNOME 3.20 the behavior changed in the C library. It seems it should replace every parameter with null. Maybe for a bug the python library does not do this correctly, I'm not sure.

mattrose:
Works. Pulling.  Thank you so much!

PR #435, opened 2021-05-07
Added set_tab_title command to remotinator.

yusufgungor:
I have added **setting tab title** function to remotinator. I think it will be useful for others too.

Thanks all contributors for this great project.


mattrose:
Thank you!

Issue #433, opened 2021-05-04
Ratio and/or position on a config file's layout ignored


facundofc:
Ratio and/or position properties on a VPaned type node on a layout are not obeyed when starting up.

To reproduce, open a terminator window. Split it horizontally or vertically. Resize one pane to fill the entire window. Left-click on the window and Preferences. Go to Layout tab and click on the Add button. Then close the preferences dialog. Run a terminator session recalling the just saved layout (say it's named A):

```
terminator -l A
```

Note that there is a new window with two panes, but equally sized.

The expected behavior is, I believe, to have the exact same layout as the saved one. This is, one pane occupying the entire window.

I'm running Ubuntu Focal:

```
$ uname -a
Linux kathy 5.4.0-72-generic #80-Ubuntu SMP Mon Apr 12 17:35:00 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 20.04.1 LTS
Release:	20.04
Codename:	focal
```

mattrose:
fixed by PR #442

PR #442, opened 2021-06-01
Fixed issue #433 (layout sometimes ignores ratio)

Vulcalien:
Since the bug is well explained in the issue, let's talk about the patch.

I studied this problem for a few hours.
Apparently, this bug *always* occurs when there are two terminals (horizontal or vertical, doesn't matter).
It also occurs when there are more than two, but I didn't figure out any logical pattern.

### Technical details
Basically, while loading a layout, the function `set_position_by_ratio` indirectly calls `new_size` (through Gtk.main_iteration()) which calls `self.set_position(self.get_position())`.
Only after set_position is executed the position is actually set (inside `set_position_by_ratio`), based on the ratio.
But the problem is that set_position takes self.get_position() as argument, while it still hasn't been set. So it calculates the new ratio, which is always about 0.5.

### Fix
Initially, I wanted to remove `self.set_position(self.get_position())` from `new_size`, but that would break stuff. (When splitting a terminal, the ratio has to be calculated again because of that line of separation between the terminals).
So I instead just store self.ratio on a temp variable and restore the value after Gtk.main_iteration() has done.

### DIFF
--- a/terminatorlib/paned.py
+++ b/terminatorlib/paned.py
@@ -488,11 +488,15 @@ class Paned(Container):
         return float(position) / float(non_separator_size)
 
     def set_position_by_ratio(self):
+        # For reasons unknown to me, self.ratio often changes when the following loop is executed
+        ratio = self.ratio
+
         # Fix for strange race condition where every so often get_length returns 1. (LP:1655027)
         while self.terminator.doing_layout and self.get_length() == 1:
             while Gtk.events_pending():
                 Gtk.main_iteration()
 
+        self.ratio = ratio
         self.set_pos(self.position_by_ratio(self.get_length(), self.get_handlesize(), self.ratio))
 
     def set_position(self, pos):


mattrose:
Thank you so much!  I must admit I looked at that for a little while when it came in and it was not immediately obvious.  I'll pull this down this evening and give it a once-over but you've explained very clearly what's gone wrong and how you fixed it, so I'm sure it'll get merged in very soon.

Thanks again!

mattrose:
Works brilliantly.  This seems to be the second hack around inconsistent GTK calls, and now I think I know why they're getting rid of the GTK.Container() class in GTK4.  It does mean that paned.py will need to be re-written from scratch when we go to GTK4, but that will probably be a good thing.

mattrose:
Any other bugs you feel like fixing? 😃

Vulcalien:
Sure! I've plenty of time, I'll try to fix some stuff

PR #443, opened 2021-06-02
Bug Fix: 'Clear selection on copy' is always unchecked

Vulcalien:
The checkbox for "Clear selection on copy" was always unchecked, because the widget active status was not set.
And the id of the checkbox in preferences.glade was not set.

### To Reproduce
Try to check "Clear selection on copy" and then close the preferences menu. When opened again, it's unchecked.

### DIFF
--- a/terminatorlib/preferences.glade
+++ b/terminatorlib/preferences.glade
@@ -790,7 +790,7 @@
                                   </packing>
                                 </child>
                                 <child>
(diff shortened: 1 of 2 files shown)


mattrose:
Thanks again for all your help.  I'm really busy with other life aspects right now and this really does help, both the project and me mentally.  Nothing makes me happier than seeing an unbidden PR that fixes a bug.

PR #454, opened 2021-06-22
Incorrect layout sorting in notebook

### LINKED ISSUE #453: Incorrect layout sorting in notebook
## Description
Layouts are in incorrect order inside the notebook if the layout count is above 9 in the notebook.

## Steps to Reproduce
1- Create a layout, add a window in it. 
2- Add a notebook inside the window.
3- Add more than 9 child layouts ( preferably of type terminal ) inside it. 
4- Give a title to each terminal layout in order to differentiate between them.

## Expected behavior
All the layouts should be in the specified order.

## Reason
After going through the code, the problem is the compare child function inside the notebook

AsadJivani:
closes #453

### DIFF
--- a/terminatorlib/notebook.py
+++ b/terminatorlib/notebook.py
@@ -78,8 +78,8 @@ class Notebook(Container, Gtk.Notebook):
     def create_layout(self, layout):
         """Apply layout configuration"""
         def child_compare(a, b):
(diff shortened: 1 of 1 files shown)


mattrose:
Thank you so much.

PR #456, opened 2021-06-25
Fixed Issue #425 (hide_window will try to show a destroyed window)

Vulcalien:
## The problem
The function `on_hide_window` is called even if the window was destroyed.

## More details
On line 135 we bind that function using `Keybinder.bind(...)`, but never "unbind" it. Now, since all the terminator windows are run by one python process, this key binding remains until all windows are closed. So what happens is that when the key combination for hide_window is typed, `on_hide_window` is called even if the window was destroyed.
The first thing I tried to do was to unbind it, but this caused all the windows to lose the binding, so that hide_window would not work anymore at all.

## Solution
Simply store in a variable `isDestroyed` the value True when the window is destroyed, and do nothing if `on_hide_window` is called.

### DIFF
--- a/terminatorlib/window.py
+++ b/terminatorlib/window.py
@@ -40,6 +40,7 @@ class Window(Container, Gtk.Window):
     title = None
     isfullscreen = None
     ismaximised = None
+    isDestroyed = False
     hidebound = None
     hidefunc = None
     losefocus_time = 0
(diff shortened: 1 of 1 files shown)


mattrose:
This looks good, and might fix a lot of niggling issues that I haven't been able to track down.  I'll pull it down and take a look at it, but assuming everything is fine I'll likely merge it in tonight.

PR #459, opened 2021-06-29
Add new plugin for opening current directory using right mouse button

pktiuk:
Hello

I have created this simple plugin to add a feature I really missed in terminator.  

Simple opening current directory.

obraz

### DIFF
--- a/terminatorlib/plugins/dir_open.py
+++ b/terminatorlib/plugins/dir_open.py
@@ -1,0 +1,32 @@
+from gi.repository import Gtk
+
+from terminatorlib.config import Config
+from terminatorlib.terminal import Terminal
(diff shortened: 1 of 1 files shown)


pktiuk:
I think this plugin could be made enabled by default (after some refinement).

mattrose:
I love it.  It's very simple and could come in handy.

pktiuk:
I think it would be worth adding a  translation to this addon and enable it by default.  
I will later open separate PR-s for this.

pktiuk:
@mattrose   
When can we expect a new release of terminator?  
The previous one is from April 2021 and there are many improvements which should be shipped to end-users.

PR #460, opened 2021-07-02
Update translation

pktiuk:


mattrose:
Thanks for this.  I know this is the grindy, non-fun bit 😄

PR #461, opened 2021-07-02
Enable dir_open plugin by default

pktiuk:
I think it may be a good idea to enable this plugin by default, it may be especially useful for new users not aware of the existence of plugins. 

obraz


mattrose:
I'm *really* wary of adding stuff to the right-click menu by default.  It's already too long and busy IMO.  I think plugins is one of the more obvious features of terminator, so I think that people will find it if they need it.

pktiuk:
Okay,  
I am closing to avoid clutter

mattrose:
Thanks for understanding.

Issue #400, opened 2021-03-05
Terminal loses focus versus tab title


aplopez:
**Describe the bug**
When clicking on the current tab's title, the terminal loses the focus which is transferred to the title. It is no longer possible to type on the terminal.
Running `terminator -g /dev/null` still behaves as described.

**To Reproduce**
1 - Open at least 2 tabs.

2 - Select a tab by clicking on it and you can type your commands on that terminal.
![Step-2](https://user-images.githubusercontent.com/17848382/110122155-89429600-7dbf-11eb-9aa8-45fe23037246.png)

3 - Click (again) on current tab's title as you just did on step 2.

4 - The terminal loses the focus (it is slightly greyed and the red bar become grey), which is transferred to the tab's title (see the dotted-line box around the title) and you can no longer type on the terminal.
![Step-4](https://user-images.githubusercontent.com/17848382/110122176-9069a400-7dbf-11eb-9175-c38d6bdd026a.png)

5 - Click on the terminal itself, and the focus is transferred from the tab's title back to the terminal. You can again type your commands.

**Expected behavior**
When clicking for a second time (not a double-click) on the tab, the focus should remain on the terminal. This is what seems most logical and what Gnome Terminal does.

**Desktop (please complete the following information):**
 - RHEL7

**Additional context**
Seems to be this same [old bug on Launchpad](https://bugs.launchpad.net/terminator/+bug/1520692) which was never fixed.

mattrose:
So, I think that when you click on the same tab that's associated with the terminal, it transfers focus to the tab itself, not the associated terminal.  I wonder if there's a way to associate just a double-click to the associated tab, and not a single-click.   I may see how gnome-terminal does this.

mattrose:
This is the gnome-terminal fix, apparently:
https://bug746665.bugzilla-attachments.gnome.org/attachment.cgi?id=300270

Not sure how to translate that to python.

marktimarev:
> This is the gnome-terminal fix, apparently:
> https://bug746665.bugzilla-attachments.gnome.org/attachment.cgi?id=300270
> 
> Not sure how to translate that to python.

Thanks, I will try to figure out how implement it

marktimarev:
Yeah, I got it, they hook the event when a tag grabs focus and then push the focus back to the inner terminal back manually.
Going to push a MR soon.

mattrose:
Wow, that's great!  Thanks and looking forward to the PR.

[...]

PR #470, opened 2021-07-22
Set CAN_FOCUS to False for notebook widgets

marktimarev:
Resolves [#issue400](https://github.com/gnome-terminator/terminator/issues/400)

Apparently, we do not need to subscribe on focus-grub event and put the grab back to the VTE terminal. Adding a setting can_focus=false is enough.

Applying it I am still able to rename the tab by a double click, though just clicking on it does not steal l the focus.

### DIFF
--- a/terminatorlib/notebook.py
+++ b/terminatorlib/notebook.py
@@ -39,6 +39,8 @@ class Notebook(Container, Gtk.Notebook):
         self.connect('scroll-event', self.on_scroll_event)
         self.configure()
 
+        self.set_can_focus(False)
+
         child = window.get_child()
         window.remove(child)
         window.add(self)


mattrose:
This looks perfect and I'm sure it works, but I'll just take a look at this and merge this tomorrow (Friday) if work is not too busy.  Thanks so much!  It's hard to believe a bug that's been open for 4 years has a one-line fix.

marktimarev:
Well, well. We still need your check in the sense that the commit made in gnome-terminal at https://bug746665.bugzilla-attachments.gnome.org/attachment.cgi?id=300270 clearly states that:

> Merely setting TerminalTabLabel:can-focus to FALSE is not enough to
> prevent the keyboard focus from entering the tab label.

That's exactly what I did but it works for me. Maybe there was another problem in GTK at the moment that was fixed during these 4 years, and so now we can just use `can_focus=False`

mattrose:
Works for me too, merging!

mattrose:
Thanks again

Issue #446, opened 2021-06-05
Launch a new command in a vertical or horizontal split.


zachriggle:
**Is your feature request related to a problem? Please describe.**
Terminator lacks feature of other terminal emulators like `tmux split-window -h zsh` which splits the existing window horizontally.

**Describe the solution you'd like**
It would be useful if there were a way to launch a program in a vertical or horizontal split of the current, EXISTING window.

**Describe alternatives you've considered**
Layouts are useful, but one can only launch a new window with a new layout, not modify an existing window/layout.

**Additional context**
"Super Guide to the split-window tmux Subcommand (and Beyond)"
https://gist.github.com/sdondley/b01cc5bb1169c8c83401e438a652b84e

zachriggle:
Update: I recently learned about `remotinator` which does allow creating new splits, but does not allow specifying the command run within them in the way that tmux does.

davewongillies:
I think it'd be useful if the functionality of `remotinator` was merged into `terminator`. Its kind of weird that `remotinator` exists when `terminator` itself has the `--new-tab` switch, which is pretty similar to alot of what `remotinator` does.

mattrose:
Agreed.  I always thought it was weird, but it was like that when I got here :smiley: and I'm still trying to figure out the best way to re-incorporate it into the main terminator script.

davewongillies:
> Agreed. I always thought it was weird, but it was like that when I got here  and I'm still trying to figure out the best way to re-incorporate it into the main terminator script.

I looked into this back when terminator was hosted on launchpad but didn't get anywhere since I couldn't figure out how to contribute.

I've noticed that there's already functions for creating horizontal and vertical splits, eg: 

https://github.com/gnome-terminator/terminator/blob/master/terminatorlib/ipc.py#L141-L149

So it shouldn't be too hard. If I have the time I might take a crack at it since it looks like the bulk of the work might already be done.

mattrose:
PRs are always welcome.  One of the reasons we moved to github was to encourage more people to contribute. I had more free time last year, but now both Markus and  I are both doing this in our spare time and appreciate all the help we can get.

mattrose:
Had some free time today and implemented this; You can now run *split [-T <title>] [-x <command>]` and it will split the termnal and run a command in the new terminal

PR #472, opened 2021-07-25
add parameters to remotinator split commands

mattrose:
This is to add parameters to the remotinator vsplit and hsplit command.  You can now run `*split -T <title> -x <command>` and it will split the termnal and run a command in the new terminal


mattrose:
fixes #446

evs-xsarus:
@mattrose when do you expect a new release with this feature?

PR #480, opened 2021-08-06
Issue 365

### LINKED ISSUE #365: Broadcast to ... "This Window" and "This Tab"
Hey guys. Love this terminal emulator and have been using+recommending it forever.

I like to use the "Broadcast {x}" functionality. But it's slightly slow in certain situations. For example, sometimes I may have many terminals open over many workspaces, and I don't always want to broadcast to every terminal window. Sometimes I'd like to only broadcast to every terminal pane embedded inside the current window. Other times I'd like to broadcast only to terminals residing (via splits) in the current window's tab.

To solve this problem, I use Broadcast to group. But then I have to manually set each pane to the same group, which can consume some extra time, needlessly.

It would be great to have two new broadcast options:
1. *Broadcast to Window*, which would broadcast to all tabs and split panes inside the same terminator window.
2. *Broadcast to Tab*, which would broadcast to all split panes inside the same terminator tab.

Broadcast to groups works for this purpose, but with extra clicks making sure each pane has the correct group setup.

**Additional context**
To clarify the terms I'm using. Window on the left has 2 tabs, and 2 split panes on the visible tab. Window on the right has 4 panes on the visible tab.

image

mattrose:
Implements the "Group all in window" feature that was requested.

Fixes #365


mikeperalta1:
Oh man, I can't wait.

Issue #379, opened 2021-02-10
Feature request: move titlebar colors config to profiles, rather than global settings


tronorama:
When I use completely different color scheme profiles for day and night, one titlebar color scheme ends up looking jarringly out of place with one of these profiles

mattrose:
This *should* be fairly easy, but it might be trickier than just moving the titlebar config in the interface.  Very good idea, thank you!

Vulcalien:
Working on this, a PR should be there tomorrow. All the settings can be moved, except 'Titlebar at bottom'.

PR #481, opened 2021-08-08
Move titlebar settings to profiles (#379)

Vulcalien:
This PR is about the issue #379.
Since the titlebar settings are mostly about colors and fonts, it would make sense to have different settings for different profiles.

Pros of this:
- it's easier to distinguish between terminals; just look at the titlebar color
- the "Global" interface contains less stuff

Cons:
- if you want to use only a color theme for all profiles, you'll have to set the titlebar settings once per profile

## What I did
First, I wanted to test whenever it was possible or not to move those settings. I found out it is possible, and I had to do very little changes to the code.
The only one that cannot be moved is "Titlebar at bottom", because there is some code that adds it when the terminal is created (however, I think it might be possible).

Then, I went to fix the GUI.

The second commit just updates `preferences.glade` to the format used by Glade 3.38.2. (note 1)

The third commit contains all the GUI changes. Basically, I just copy-pasted some of the components from the Global menu to the new Profiles>Titlebar menu.
I chose to put the titlebar menu at the end, just to avoid moving the other ones, that I think would confuse the long term users.
The "titlebar at bottom" setting is now inside the Appearance group of the Global settings.

Also, I fixed some of the grids in the GUI, that had empty rows or columns. This should not affect the GUI and I didn't notice any difference while reviewing it.

Last thing, the minimum window size was reduced. (note 2)

image

This is the Titlebar menu
image

## Notes
1. I've noticed that the file was already been upgraded to 3.36 recently, but then was downgraded to 3.22. Now it's back to a recent version.
2. Since the minimum width was reduces, it's harder to use some of the sliders. Maybe I will try to change the position of those in the GUI.


mattrose:
Thanks so much for this.  I'm having a down day today but I'll look at this soon.

What happens when you have a config file that specifies the titlebar colors in the old place and you upgrade?

Vulcalien:
Don't worry, I'll not change the PR.
About the old config, I actually didn't think of that, going to test it right now.

Vulcalien:
It seems to just delete the old settings once terminator writes the config file, as expected.

However I just found a little bug in the GUI, that doesn't update some of the titlebar settings when switching between profiles. I think I can fix it before the PR is merged.

mattrose:
Awesome, thanks!

mattrose:
Don't worry I'll wait for your other fix before merging.  Thanks again!

Vulcalien:
Ok perfect, it should be fine now. I had forgotten to move some "update GUI" code.
Fun fact: I didn't notice because for some reason, the titlebar colors were already updated when the profile changed, despite them being global settings.

mattrose:
This seems to do the right thing

mattrose:
Thanks again!  Much Appreciated!

Vulcalien:
Glad to help!
Now that those settings were moved to Profiles, is there any documentation to update?

mattrose:
Grr.  Documentation.  If you want to update the documentation and screenshots at https://github.com/gnome-terminator/docs/blob/master/source/preferences.rst that would be fantastic!

PR #482, opened 2021-08-10
GUI: set all CheckButtons off to avoid blinking + remove grid empty rows/columns

Vulcalien:
This is a small change to the preferences.glade file.

After the last PR (#481) I've noticed some of the checkboxes "blink".
Basically, in the glade file, they are "on", but when terminator reads the config file, they are set off. So there is this very small time span where they are on, and suddenly turn off, causing some sort of blinking.

I think the reasons because this did not happen before, is that the GUI was bigger and there was more stuff to load, so there was enought time to read and set all the settings.

So I set all checkboxes to false by default, so that they turn on only when Terminator reads the config. I also removed some empty rows/columns from that file.

### DIFF
--- a/terminatorlib/preferences.glade
+++ b/terminatorlib/preferences.glade
@@ -460,7 +460,6 @@
                                     <property name="can-focus">True</property>
                                     <property name="receives-default">False</property>
                                     <property name="xalign">0.5</property>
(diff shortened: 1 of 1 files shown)


mattrose:
Looks good, thanks!

Issue #467, opened 2021-07-13
Changing cursor foreground color


kraftwerk28:
In gnome-terminal it is possible to set both cursor background/foreground colors independently, but in terminator there's only one option.

![](https://user-images.githubusercontent.com/31807671/125505627-75af7915-5711-484e-8f72-19953b14c63f.png)
*gnome-terminal settings*

![](https://user-images.githubusercontent.com/31807671/125505753-0c862b98-03aa-47e4-808c-f3917206c7d1.png)
*terminator settings*

Could it made so that terminator also has options for cursor fg/bg?

mattrose:
yes.

mattrose:
fixed by #486

PR #486, opened 2021-08-14
Cursor: make it possible to change foreground color (#467)

Vulcalien:
There were some changes to the GUI and to the config file.
Let's talk about the last one first.

## Config
Currently, there are only two settings for the cursor colors: `cursor_color` and `cursor_color_fg`. Their names can easily confuse, because `cursor_color` is the background color, while `cursor_color_fg` is a boolean (False = use a custom background; True = use the foreground as the background).
I removed them both, and replaced them with three new settings: `cursor_fg_color`, `cursor_bg_color` and `cursor_color_default`. The first two are color strings, foreground and background. The last is a boolean (False = custom colors, True = default colors).

## Python code
Most of the changes are made to make the new GUI work.
I removed the old signal functions and the old logic that initializes the cursor colors in the preferences menu, replacing everything with the new code.

## GUI
I changed the GUI related to the cursor, in the `Profiles > Global` tab.

I also fixed some inconsistencies that I noticed in the whole menu:
- "Text color" and "Font color" are now "Foreground"
- On the `Profiles > Titlebar` the Font selector was missing the "indentation"
- "Background color" in `Profiles > Colors` is now just "Background"


mattrose:
Working on another feature right now, but I'll take a look at this by monday hopefully

mattrose:
Wow, this was a lot more work than I thought it was.  Great job, and thanks again.

PR #493, opened 2021-08-24
remove gtk-update-icon-cache from setup.py

### LINKED ISSUE #102: Remove gtk-update-icon-cache handling in setup
I don't see a reason why we should generate the cache like this, its even bad for packages. So (package) maintainers should disable it.

Debian:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=959856

mattrose:
fixes #102


mattrose:
note to @lazyfrosch and @dmaphy and other package maintainers.  If we're doing any special handling of the icon-cache in our package files, we should take that out.

PR #494, opened 2021-08-25
Make tabs detachable + minor bugfix

### LINKED ISSUE #302: Feature Request: detach tab
Sometimes I want to detach a tab, similarly to what I can easily do on Firefox. I can't find a way to do that easily on Terminator, but I can do that in a boring way:

- `ctrl+shift+N` will open a new window for me.
- Then I can drag-n-drop all the panes I need from the old window to the new
- then I can remove the original pane created in the new window

which works! But, well, something more intuitive would be very good to have.

Vulcalien:
Fixes #302

Note - This pull request is to show the progress. I have done very little testing, so I don't recommend pulling it right now.

## About
As said in the issue comments, `set_tab_detachable` does the job. The [code](https://stackoverflow.com/a/56377756) found on StackOverflow gave me a guideline, but much of that stuff is already implemented somewhere in Terminator. So I had to create only one function, `create_window_detach`.

About `self.hoover`: without calling that function, if a tab is detached and a notebook only has 2 tabs, the resulting notebook only has 1 tab, causing some troubles.

Most of the work was about reading the documentation.

---

I'll finish testing soon, and maybe fix that Drag-and-drop problem that we were talking about in the comments, since it's related.

### DIFF
--- a/terminatorlib/notebook.py
+++ b/terminatorlib/notebook.py
@@ -88,6 +88,7 @@ class Notebook(Container, Gtk.Notebook):
         window.resize(size.width, size.height)
 
         self.detach_tab(widget)
+        self.disconnect_child(widget)
         self.hoover()
         window.add(widget)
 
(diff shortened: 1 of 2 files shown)


mattrose:
Is this ready to go?

Vulcalien:
Yep, I think it is.

mattrose:
Cool, I'll pull it down again and give it a quick once-over before merging it.

mattrose:
This looks great.  Thanks again!

PR #499, opened 2021-08-30
Add a "Copy" button to clone profiles

Vulcalien:
Very simple changes, but such a big time saver.
Currently, if you want to duplicate a profile, you either do it in the config file, or you have to add a new profile and manually set all of the settings again.

Also as a note, I think the `Disable Ctrl+mousewheel zoom` button in Profiles is broken. It sometimes does not change value when switching between profiles.

### DIFF
--- a/terminatorlib/config.py
+++ b/terminatorlib/config.py
@@ -310,6 +310,10 @@ class Config(object):
         """Get our profile"""
         return(self.profile)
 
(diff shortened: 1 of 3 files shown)


mattrose:
Sorry, forgot about this.  I'll look at this right now

mattrose:
Works great as always.  Thanks!

Issue #498, opened 2021-08-30
[Feature request] Too hasty on coffee protection


nickma82:
**Is your feature request related to a problem? Please describe.**
Terminator is my Nr.1 terminal frontend. I'm using it for hours every day and I never ever close it unless I want a new kernel getting loaded.
However, from time to time (like a couple of minutes ago) I accidentally close it; I immediately go "NOOOOO". And I which for something I'd never wish for any other app: 
A prompt before closing. 

**Describe the solution you'd like**
A prompt before exit, with the ability to cancel it.

**Describe alternatives you've considered**

- Unlinking/Rebinding all system shortcuts for closing apps
- Implementing a trap on bash/zsh exit 

Best

Vulcalien:
Some of the features you are talking about are already implemented.

If there is more than one terminal, Terminator will prompt you if you really want to close multiple terminals.
If you don't see this happening, go to `~/.config/terminator/config` and delete this line:
```python
suppress_multiple_term_dialog = True
```

Now, if you only have one terminal open, there is an hack you can do to make Ctrl+D not close the terminal.
I guess you want only one terminal to be kept alive (allowing the ones you create by splitting to be closed).

# Make Ctrl+D not close the terminal
## Editing the config file
If you prefer editing the config file, then open it and duplicate the `[[default]]` profile (or the name of the profile you use). Now you should have something like this:
```python
[profiles]
  [[default]]
    ...
    ...
  [[default]]
    ...
    ...
```

Change the name of the second profile to `dont-close` or similar and add three lines at the end of it:
```python
exit_action = restart
use_custom_command = True
custom_command = zsh
```

Now go to preferences, Layout. Select your layout (probably `default`). Select the Terminal in the list and change its Profile to dont-close.
And... done! Now the first terminal you open will not be closed by Ctrl+D

## Doing the same using the Preferences menu
Open preferences, go to Profiles, Add a new one.
On the new profile, go to the Command sub-menu, check `Run a custom command...` and write the name of your shell. Also change the value of `When command exits` to `Restart the command`.

Now go to preferences, Layout. Select your layout (probably `default`). Select the Terminal in the list and change its Profile to dont-close.

By using this method, you also have to manually copy the settings from the old profile to the new one.

---

That's just an hack, not a "clean" way of doing it, but it should work for Ctrl+D at least.
To close Terminator, Alt+f4 or the X button will still work without a prompt.
Coding it should not be too hard for us.

mattrose:
[...]

PR #500, opened 2021-08-30
BugFix: terminal won't restart if there is no custom command

Vulcalien:
Fixes a problem that we discovered in #498, but not the entire issue.
For some reason, something related to debug makes Terminator crash. That might be another bug.

Apparently, the parameter order was messed up, and the `Restart the command` thing would work only when there is a custom command. With this patch it should work fine even if it's not selected.


mattrose:
will pull when checks complete.

mattrose:
Thanks again! :)

Vulcalien:
You're welcome! This was a quick one.

Issue #427, opened 2021-04-21
hide_window key on Arch Linux with GNOME 40 doesn't work


Darkproduct:
I recently started to use arch Linux with gnome 40 and transferred my terminator config from Ubuntu to arch.

The keybindings where the first thing I had to fix. Instead to `<strg>` in arch you have to use `<ctrl>` for some reason and if I edit the config inside the preferences window, terminator will sill save the not working `<strg>`.

But that's not the end of the world, that really annoys me is the not working `hide_window`. I usually use the `Insert`-Key, but every other combination, like `<ctrl>j`, doesn't work.

All the information:
```
❯ neofetch
                   -`                    swt@swtarch 
                  .o+`                   ----------- 
                 `ooo/                   OS: Arch Linux x86_64 
                `+oooo:                  Host: VMware7,1 None 
               `+oooooo:                 Kernel: 5.11.14-arch1-1 
               -+oooooo+:                Uptime: 36 secs 
             `/:-:++oooo+:               Packages: 760 (pacman) 
            `/++++/+++++++:              Shell: zsh 5.8 
           `/++++++++++++++:             Resolution: preferred 
          `/+++ooooooooooooo/`           DE: GNOME 40.0 
         ./ooosssso++osssssso+`          WM: Mutter 
        .oossssso-````/ossssss+`         WM Theme: Adwaita 
       -osssssso.      :ssssssso.        Theme: Adwaita [GTK2/3] 
      :osssssss/        osssso+++.       Icons: Adwaita [GTK2/3] 
     /ossssssss/        +ssssooo/-       Terminal: terminator 
   `/ossssso+/:-        -:/+osssso+-     CPU: AMD Ryzen 5 5600X (6) @ 3.699GHz 
  `+sso+:-`                 `.-/+oso:    GPU: 00:0f.0 VMware SVGA II Adapter 
 `++:.                           `-/+/   Memory: 1040MiB / 7922MiB 
 .`                                 `/
                                                                 
                                                                
❯ showkey -a

Press any keys - Ctrl-D will terminate this program

^[[2~ 	 27 0033 0x1b
 	 91 0133 0x5b
 	 50 0062 0x32
 	126 0176 0x7e
^D 	  4 0004 0x04

❯ cat .config/terminator/config
[global_config]
  window_state = fullscreen
  always_on_top = True
[keybindings]
  split_horiz = <Ctrl>h
  split_vert = <Ctrl>v
  hide_window = Insert
[...]

PR #510, opened 2021-09-11
Bug Fix: hide_window keybinding ignores Shift key

### LINKED ISSUE #509: hide_window mapped with Shift+Control+Alt+$an_alphabet catches Control+Alt+$an_alphabet instead
**Describe the bug**
If the keymapping of `hide_window` is set to `Shift+Control+Alt+$an_alphabet`, the keystroke will be ignored and`Control+Alt+$an_alphabet` will be made to trigger hide_window instead. This occurs with (but most likely not limited to) the default `Shift+Control+Alt+A` and `Shift+Control+Alt+H`.

The problem appears to be `hide_window` specific. It does not occur with e.g. `layout_launcher`.

For the record, changing `edit_tab_title` away from `Control+Alt+A` (its default) does not make the problem go away.

 - OS or Linux Distribution: Arch Linux
 - Terminator Version 2.1.1

Vulcalien:
Fixes #509

The shift key gets ignored by the global Keybinder. That's not a bug in the library, it's a misconfiguration.
By default, keybinder uses "cooked accelerators".

I quote from the issue comments.

> The best documentation I found is the comments in the code. They are pretty good.
>From [this](https://github.com/kupferlauncher/keybinder/blob/master/libkeybinder/bind.c):
> > "Cooked" accelerators use symbols produced by using modifiers such as shift or altgr, for example if "!" is produced by "Shift+1".

There is this C function
```c
keybinder_set_use_cooked_accelerators
```

So yeah, we just had to disable them using this:
```py
Keybinder.set_use_cooked_accelerators(False)
```
---

There is another issue about hide_window not working (#427). I doubt it's related to this, I think that one is about Wayland.

### DIFF
--- a/terminatorlib/window.py
+++ b/terminatorlib/window.py
@@ -28,6 +28,7 @@ if display_manager() == 'X11':
         gi.require_version('Keybinder', '3.0')
         from gi.repository import Keybinder
         Keybinder.init()
+        Keybinder.set_use_cooked_accelerators(False)
     except (ImportError, ValueError):
         err('Unable to load Keybinder module. This means the \
 hide_window shortcut will be unavailable')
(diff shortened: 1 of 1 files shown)


mattrose:
thanks!  Merging

Issue #513, opened 2021-09-16
Auto-scroll to the cursor position when typing


marktimarev:
In the Terminal, if you scroll up and then start typing, then the terminal automatically scrolls down to the cursor position. It's really convenient since I do not need to scroll down myself every time I checked some long output and want to issue another command.
Since I started to use Terminal I miss that feature and trying to keep in mind I need to scroll down before typing.

mattrose:
Interesting, I went to look at your PR and this works as described, even before applying your PR.  Can you be a little bit more specific about the environment you find yourself having this problem?

marktimarev:
Hmm

Well, I just checked out to the current HEAD (before my commits). Then I run the app with `python3 terminator`.
As the window appear, I hold the enter key for a while to make many lines. Then, I scroll up so that the last line is not visible. And here is the problem: if type something, then nothing happens at all, and if I manually scroll down to the last line I will see characters appeared as the result of my typing. So, what I want is that if I type and the last line is not visible, it should scroll down automatically.

Does it really work as expected for you? Maybe it's something about the version VTE or some another package on my machine then..

marktimarev:
![example](https://user-images.githubusercontent.com/86591723/133657384-fd9196cd-fb6a-4a96-aa53-852181abfb1a.gif)

mattrose:
Sorry to say it really does work as expected for me.  It could be VTE, or window manager, or something else I haven't thought of.

marktimarev:
Ok, let's close it then, seems I just got unlucky.

PR #514, opened 2021-09-16
#513 Add auto-scroll down to the cursor position

marktimarev:
Resolving https://github.com/gnome-terminator/terminator/issues/513

### DIFF
--- a/terminatorlib/terminal.py
+++ b/terminatorlib/terminal.py
@@ -917,6 +917,8 @@ class Terminal(Gtk.VBox):
             dbg('Terminal::on_keypress: Called on %s with no event' % widget)
             return False
 
+        self.scrollbar.set_value(self.vte.get_cursor_position()[1])
+
         # FIXME: Does keybindings really want to live in Terminator()?
         mapping = self.terminator.keybindings.lookup(event)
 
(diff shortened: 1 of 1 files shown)


marktimarev:
I propose to watch for the event of cursor movement (it moves when you type or just manually move it with arrow keys). And it it moves to scroll at its position.
Works well for me.

mattrose:
Looks good, I'm just gonna check this out and run a sanity test before merging.

marktimarev:
Oh, I just understood that it will not work properly if there's some task running and its output, printing into the terminal, moves the cursor. I mean, if I run something in the background and want to scroll up to check something, the scroll will jump down to the cursor as the task output something.

At least, the original GNOME Terminal behaves well with such a case.

We need to somehow check that cursor movement is caused by pressing a key. We could of course just catch key down event, save the cursor position, and on key up event check if the position is changed. But I think it's a too dirty solution :(

marktimarev:
Finally, this is what we need. The event _commit_ is "Emitted whenever the terminal receives input from the user and prepares to send it to the child process."

Now auto-scroll works exactly like in Terminal.
Please check

mattrose:
Thanks for updating, I was just about to start doing some research.  I'll pull it down and check rt now

PR #516, opened 2021-09-19
Automatically focus the Keybindings menu

Vulcalien:
Fixes a bug in Preferences, Keybindings.

Right now, to activate the search bar, we need to click the tree menu.
This change makes the search bar work without having to click the widget, giving the focus to it automatically.

### DIFF
--- a/terminatorlib/preferences.glade
+++ b/terminatorlib/preferences.glade
@@ -3719,6 +3719,7 @@
                   <object class="GtkTreeView" id="keybindingtreeview">
                     <property name="visible">True</property>
                     <property name="can-focus">True</property>
+                    <property name="has-focus">True</property>
                     <property name="model">KeybindingsListStore</property>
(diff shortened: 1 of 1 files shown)


mattrose:
AFAICT the search window is already focused when you bring it up in Preferences->Keybindings

Vulcalien:
I've just tested it on two VMs. One is Debian 11 + dwm. The other one is Debian 10 + GNOME (terminator version is 1.91). Both use Xorg.
Same behavior, the search bar never appears if I try to type. I have to click inside the widget, and then it works.

Might it be Xorg?

---

Also, I've noticed that the other TreeViews (like Preferences>profiles and layouts) automatically focus the widget, without having to specify the "has focus" property. So I don't really understand why this happens.

mattrose:
I got it now.  I figured out what you wanted to do.

Issue #518, opened 2021-09-26
Feature: Configuration to colorize split screens from default grey colour, Issue: Remove ability to select 0 and 1 handle_size from configuration gui


caprinux:
Thank you to the devs and the community for continually updating and keeping terminator alive, it has definitely been one of the best terminal that I've used and it definitely has much more potential to grow.

As a long time user of terminator, I've always been rather troubled by the **handle_size** or **border_size** of split screens in terminator that couldn't seem to get any smaller than the supposed `-1` that is in the configuration.

I even tried manually modifyign `handle_size` in my terminator config to even smaller values but to no avail this thick looking split screen border still seems to be unable to get any smaller.

![image](https://user-images.githubusercontent.com/76640319/134811546-efa6506b-9fa8-4687-8408-9028b50d45c6.png)

It would be great if we could manually modify and reduce the handle_size even further, to be just a few pixels wide, which would look so much more aesthetically pleasing. 

I may make a PR on this myself if I have the time to, so if someone is unable to work on this and have any suggestions on which files I should look at, please let me know as well!

Additional Feature Ideas:
Maybe even a color option for handle_size could be possible :)?

caprinux:
Realised that `1` handle_size was exactly what I saw looking for after reading the code. Would be better if `-1` and `0` weren't provided as options inside of the settings gui though

caprinux:
Issue: Remove `-1` and `0` from handle_size configuration panel.

Feature: Colorize split screen from default grey color

PR #519, opened 2021-09-26
Fix terminal separator size setting

caprinux:
Partially addresses #518 , not sure where I can insert a config to modify color of terminal seperator so I'll put that one on hold for now.

1. Make it such that if handle_size is out of range, handle_size will be defaulted to smallest.
2. Terminator seperator size in gui settings will have range from 1 to 20, instead of -1 to 20 which is extremely misleading as -1 and 0 is out of range and previously defaults to a larger value.


mattrose:
Thanks for this.  I'll check it out.

PR #520, opened 2021-10-01
Add paste_selection keybinding.

rkitover:
Unbound by default, pastes primary selection.

Signed-off-by: Rafael Kitover <rkitover@gmail.com>


mattrose:
Personally, I just use middle mouse click to paste the primary selection, or Ctrl-Shift-V to paste from the clipboard.  I'm assuming you don't have a middle mouse button?

rkitover:
I do have a middle mouse button, but I prefer sometimes to use a keyboard shortcut for this. QTerminal supports this feature. Using a keyboard shortcut can be less error prone, e.g. if your middle button is the scroll wheel.

PR #525, opened 2021-10-12
Allow multiline commands in Custom Commands plugin

VDuchon:
Custom Commands plugin was able to save only one-line commands. This enables to save and execute commands with multiple lines.


mattrose:
works for me, thanks, and sorry about the wait.

Issue #534, opened 2021-11-03
Vte.Terminal().set_encoding deprecated


mattrose:
We should remove the encoding section from the context menu and the "Profiles->Compatibility" tab, as support for different encodings has been deprecated, and only UTF-8 is supported

pkreuzt:
Sorry to necrobump this issue, but I'm confused about it. Gnome-terminal, Xfce4-terminal and other terminals using VTE (2.91 in my distro) still support multiple encodings. They don't support fast change via context menu as in Terminator, but they never did. I've been reading VTE's Gitlab and didn't find references to this. Am I missing something ¿!?

mattrose:
According to docs this has been deprecated for years now.  https://lazka.github.io/pgi-docs/#Vte-3.91/classes/Terminal.html#Vte.Terminal.set_encoding

pkreuzt:
It hasn't been removed from VTE' master (yet). Then, everyone else is relying on a doomed function? The recommended luit workaround doesn't allow to change encoding on a running shell on the fly. This seems to be important for a number of users, according to issues and comments here.

PR #536, opened 2021-11-06
Remove Encoding settings (deprecated)

Vulcalien:
Issue #534

All I did was remove code. I don't expect anything breaking, since it doesn't seem to be connected to other code.

## List of changes
- The file `encoding.py` was entirely deleted.
- The GUI widget in `preferences.glade` used to select encoding was removed, along with the signal handlers in `prefseditor.py`
- The code that generated the context menu entry (file `terminal_popup_menu.py`) was removed, along with the signal handler in `terminal.py`
- The settings `encoding` and `active_encodings` were removed

I don't know if there is a way to test this, since it's a feature removal. I just opened the menus and no error messages were being printed.


mattrose:
:heavy_check_mark: Less Code
:heavy_check_mark: Smaller  context menu

I like it!

jhalx:
How should I go about changing the terminal encoding after the option has been removed?

jhalx:
> How should I go about changing the terminal encoding after the option has been removed?

One of the biggest influencing factors of my use of the terminal is that it is very convenient to modify the terminal encoding

jhalx:
I hope to add this feature back to the built-in plugin and set it to be turned off by default. Although most people don't need this feature, it is indispensable.

> How should I go about changing the terminal encoding after the option has been removed?

I hope to add this feature back to the built-in plugin and set it to be turned off by default. Although most people don't need this feature, it is indispensable.

jhalx:
> How should I go about changing the terminal encoding after the option has been removed?

I hope to add this feature back to the built-in plugin and set it to be turned off by default. Although most people don't need this feature, it is indispensable.

Vulcalien:
The reason we removed it was that vte, the terminal GTK component, deprecated it. It would be up to them to add it back, and then we could revert this change as well.

It is very unfortunate, but maybe a command like `luit` could help.

PR #544, opened 2021-11-30
os.environ does not have LANGUAGE in Centos8.

xuezhixin:
There is no LANGUAGE available on environ of centos, so there is no way to select a new LANGUAGE, even though only English is available.

### DIFF
--- a/terminatorlib/util.py
+++ b/terminatorlib/util.py
@@ -106,9 +106,9 @@ def manual_lookup():
                 target = language
                 break
     elif 'LANG' in os.environ:
-        languages = os.environ['LANG'].split('.')
-        for language in languages:
-            if language in available_languages:
+        language = os.environ['LANG'].split('.')[0]
+        for i in range(len(available_languages)):
+            if language == available_languages[i]:
                 target = language
                 break
     return base_url % target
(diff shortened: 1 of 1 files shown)


mattrose:
According to my reading, LANGUAGE and LANG are two slightly different variables with slightly different purposes and LANGUAGE is present on other linuxes.  I'd much rather use LANG as a fallback if LANGUAGE doesn't exist, rather than replacing LANGUAGE with LANG entirely

xuezhixin:
What you said is very reasonable, so I will add Lang as a judgment. Is it appropriate?If you think so, I will continue to revise it.

mattrose:
This is ... closer, but LANG should only ever return one element.  LANGUAGE returns a colon separated list, but LANG only ever returns one language. You should be able to just use the string that split returns. so something like:
```
 elif 'LANG' in os.environ:
￼        target = os.environ['LANG'].split('.')[0]
```

mattrose:
just another Q:  What does `locale` output in centos 8?

xuezhixin:
Thank you for your advice. I have revised the language selection again to facilitate the expansion of the website language in the future. Running in my environment on CentOS 8 is zh_ CN is en after changing the English is en_US.Also on Debian, language is zh_ CN.

mattrose:
Looks good, thank you!

PR #549, opened 2021-12-07
Fix: 'None' value for keybindings breaks editor

### LINKED ISSUE #548: None isn't a valid value of keybindings config anymore
Fill out as many of these fields as you can

**Describe the bug**
`None` value in `[keybindings]` settings breaks the ability to set the keys.

**To Reproduce**
1. Set `cycle_next = ""` - and launch terminator, try to change the key binding for anything. It works as expected  of shows a dialogue `X already in use to trigger Y`
2. Set `cycle_next = None`, now it's broken. I have config from an ancient version 2 years ago, back there it was a valid value

**Expected behavior**
I'd like to have the config backward compatible.

 - OS or Linux Distribution: Arch Linux
 - Version terminator 2.1.1-2
 - Display Technology: X11

**Additional context**
```
# logs for "" value
> terminator -u -g /tmp/term.config 

** (terminator:17429): WARNING **: 11:53:19.939: Binding '<Control><Alt>a' failed!
Unable to bind hide_window key, another instance/window has it.
<window.Window object at 0x7fe96198c540 (terminatorlib+window+Window at 0x55aeffa66280)> is not in registered window list
# logs for None value
> terminator -u -g /tmp/term.config   

** (terminator:17261): WARNING **: 11:52:54.237: Binding '<Control><Alt>a' failed!
Unable to bind hide_window key, another instance/window has it.
  File "/usr/lib/python3.9/site-packages/terminatorlib/prefseditor.py", line 1719, in on_cellrenderer_accel_edited
    parsed_conf_accel = Gtk.accelerator_parse(conf_accel)
TypeError: Argument 0 does not allow None as a value
```

Vulcalien:
Fixes #548 

As explained in the issue, if a keybinding's value is set to None, it becomes impossible to remap it.
However, I've noticed this affects all the keybindings: if one of them is set to None, then the Preferences Editor won't be able to change any value.
This is because function `Gtk.accelerator_parse` (which is called to avoid keybinding duplication) does not allow None as a value.

Also, since we should consider `None` as a valid value for `hide_window`, we have to check for both `''` and `None` (last changed in #515)

### DIFF
--- a/terminatorlib/prefseditor.py
+++ b/terminatorlib/prefseditor.py
@@ -1738,10 +1738,13 @@ class PrefsEditor:
 
         accel = Gtk.accelerator_name(key, mods)
         current_binding = liststore.get_value(liststore.get_iter(path), 0)
+        parsed_accel = Gtk.accelerator_parse(accel)
 
         duplicate_bindings = []
         for conf_binding, conf_accel in self.config["keybindings"].items():
-            parsed_accel = Gtk.accelerator_parse(accel)
+            if conf_accel is None:
+                continue
+
             parsed_conf_accel = Gtk.accelerator_parse(conf_accel)
 
             if (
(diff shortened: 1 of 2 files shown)


mattrose:
Thanks again @Vulcalien!

Felixoid:
Thanks a lot, it was quite quick!

PR #551, opened 2021-12-08
Improve and optimize the code for background images

Vulcalien:
Plus, fixed code duplication and organized code.

The problem: even if the `background_type` is not `image`, terminator will still print an error saying that the file was not found (when the file was deleted). To fix this, you should go to the config file and manually delete the `background_image` entry.

Solution: check if `background_type` == `image` before attempting to read the image file.

I also removed the duplicated code, that was identical to set_background_image.

In background_draw (which by the way seems to have bad performance) becomes unnecessary to check for `background_type`.

### DIFF
--- a/terminatorlib/config.py
+++ b/terminatorlib/config.py
@@ -212,6 +212,7 @@ DEFAULTS = {
                 'background_color'      : '#000000',
                 'background_darkness'   : 0.5,
                 'background_type'       : 'solid',
+                'background_image'      : '',
                 'backspace_binding'     : 'ascii-del',
                 'delete_binding'        : 'escape-sequence',
                 'color_scheme'          : 'grey_on_black',
@@ -253,8 +254,6 @@ DEFAULTS = {
                 'autoclean_groups'      : True,
                 'http_proxy'            : '',
                 'ignore_hosts'          : ['localhost','127.0.0.0/8','*.local'],
-                'background_image'      : '',
-                'background_alpha'      : 0.0,
                 # Titlebar
                 'title_hide_sizetext'     : False,
                 'title_transmit_fg_color' : '#ffffff',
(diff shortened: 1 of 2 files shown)


mattrose:
Thank you so much for fixing this up.  I checked this background image code in in a bad state (in my defense, I wrote this while I was learning about terminator, and GTK and it was a long, frustrating time).  I can certainly merge this code as is, as it's *much* better than the code I wrote, but...

I did notice that to get the picture to display I had to restart terminator.  Is there any chance you could have the change apply as soon as the config changes?  It should be just a matter of altering Terminator().reconfigure to pick up the changes.  If not, lemme know and I'll merge the change as-is.

mattrose:
Also, any ideas you have for fixing up background_draw are welcome.  I played around a lot with this function, but this was the only way I could get it to load the image reliably on every draw.

Vulcalien:
I've noticed the need to restart. I guess there is some way to update it, just like all the other settings update.
The performance issue I think is more difficult, because I don't really see any other way of having the background other than actively drawing it, which is the slowest part.
I'll work on the "update without restarting" first.

Vulcalien:
Found a way to optimize it! And I also had the possibility to use `cProfile` for the first time.

First, using Surface instead of PixBuf we get a free performance boost.
Now, the most CPU consuming operation is the scaling of the image. By default, the 'Filter' is `GOOD`, but as we saw that's way too slow. I've tried changing the filter to `NEAREST` or `FAST` and both are perfectly fine.
We should choose one of the filters (those two or others that I haven't tried) as the default for Terminator.

Relevant:
- The patch with the FAST filter (`git diff` format): [filter.patch.txt](https://github.com/gnome-terminator/terminator/files/7688769/filter.patch.txt)
- [Documentation](https://pycairo.readthedocs.io/en/latest/reference/enums.html#cairo.Filter) for Filters.

mattrose:
Can you put the filter FAST patch in this PR as well?  It looks really good.  If people complain about image quality we can make the filter configurable later, but I'm pretty sure it'll be fine.

PR #553, opened 2021-12-10
Unzoom terminal on interaction

### LINKED ISSUE #550: Switch focus and splits don't work when terminal is zoomed
When a tab has few terminals and one is zoomed, switching to any direction doesn't work. There are different approaches in different emulators:

- tmux cancels zoom and switch to another pane
- kitty remains zoomed, but switch the pane to the next/previous one

I'd prefer tmux approach, when focus change is triggered, it's better to zoom out the current terminal and then switch

Kitty's approach would work as well. The zoom can be preserved during a terminal focus switch

**update**: the same case is for splits. It seems logical to me to exit zoom mode and split terminal. And the same for tab switching.

Vulcalien:
Fixes #550

These changes make a zoomed/maximized terminal unzoom on interactions such as: creating a new tab, resizing the terminal, splitting it etc...

There are still problems with two of these interactions: rotating the terminal and moving to another terminal.
The function `get_allocation()` returns `(width: 1, height: 1, ...)` right after an unzoom.

The bug can be replicated with a few terminals (I use one on the left, one top-right, one bottom-right). Zoom the bottom-right one and try to rotate or move to the top-right terminal.
What I get for the rotation is that one terminal is big, the other two are very small.

### DIFF
--- a/terminatorlib/notebook.py
+++ b/terminatorlib/notebook.py
@@ -278,7 +278,6 @@ class Notebook(Container, Gtk.Notebook):
                    'split-horiz': self.split_horiz,
                    'split-vert': self.split_vert,
                    'title-change': self.propagate_title_change,
-                   'unzoom': self.unzoom,
                    'tab-change': top_window.tab_change,
                    'group-all': top_window.group_all,
                    'group-all-toggle': top_window.group_all_toggle,
--- a/terminatorlib/terminal.py
+++ b/terminatorlib/terminal.py
@@ -915,20 +915,6 @@ class Terminal(Gtk.VBox):
         # FIXME: Does keybindings really want to live in Terminator()?
         mapping = self.terminator.keybindings.lookup(event)
 
-        # Just propagate tab-swictch events if there is only one tab
-        if (
-                mapping and (
-                    mapping.startswith('switch_to_tab') or
-                    mapping in ('next_tab', 'prev_tab')
-                )
-        ):
-            window = self.get_toplevel()
-            child = window.get_children()[0]
-            if isinstance(child, Terminal):
-                # not a Notebook instance => a single tab is used
-                # .get_n_pages() can not be used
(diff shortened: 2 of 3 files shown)


Vulcalien:
Now it works, but there is an annoying bug: after an unzoom, the terminal is moved by a single pixel, making the navigation after an unzoom inconsistent. It's a different issue, so I'll see if I can fix that with another PR.

mattrose:
So, I have to admit, I'm not clear on what the problem is with this issue.  That seems to be a "me" issue, as it was clear to you, but at the same time, I'm hesitant to merge something that you think is not quite complete because I have no idea :smile: .  If you could open the other PR and then I can merge them in at the same time, that'd be great.

mattrose:
I think I'm starting to figure this one out after having looked at it and read your patch, and it looks good to me.  If you could either open a new bug or come up with a new PR for the other issue you mentioned, that would be great.  That way there's some evidence of it other than in a closed PR, in case you get busy with something else, or there's something else blocking your fix.

Vulcalien:
Here is how to reproduce the issue.
1. Split the terminal
2. Zoom one of the terminal
3. Try to move to the other terminal (with keyboard shortcuts)

Current behavior:
nothing will happen, and the zoomed terminal will remain zoomed.

Desired behavior:
the terminal should unzoom and then move to the other terminal.

Other examples of "interactions" that should trigger an unzoom are a split, a rotation, a "new tab".
So with the changes in this PR, if you try to split, the zoomed terminal will first unzoom and then split as it should.

This is more or less how tmux handles the behavior as well.

***

I will make the new issue.

mattrose:
Merging.  Sorry about the wait.

Vulcalien:
Finally wrote the issue, I wanted to make sure to see the real problem. Apparently, it's focus lost.

PR #552, opened 2021-12-10
Add ability to configure cell width (font character spacing)

FernandoBasso:
Terminator already implements the VTE “cell height” feature. This PR implements “cell width”.

We also take the opportunity to refactor “line height“ to “cell height“ to match VTE terminology and to avoid having to related, similar features with two different naming style in the code base.

Thanks @mattrose for the help and encouragement <3.

terminator-cell-width

PR #561, opened 2021-12-25
Fix: handle_size treated as cell_width

Vulcalien:
In PR #552 the cell_width setting was introduced.
However, the GUI has a bug that makes the `handle_size` be considered as `cell_width`.

To reproduce:
- open preferences
- change the value of 'Terminal separator size' all the way left and then all the way right
- the 'Cell Width' setting is changing too

I also moved "Cell Height" after "Cell Width", so that we have "width, height" instead of "height, width".


FernandoBasso:
@Vulcalien Thanks a lot for spotting the problem and the fixing it! Awesome!

Issue #206, opened 2020-09-08
Create Snap and Flatpak for Terminator


mattrose:
So, there's two reasons for this:

1.  There's a lot of demand for it.
2.  There's a lot of functionality that would be nice to have in terminator, but is unavailable through the stock VTE library.  There are patches that exist, but we cannot rely on them being there.

mattrose:
If anyone has any experience creating these, PRs would be greatly appreciated.

dafanasiev:
I try to make snap... 

But i have no idea how to deal with d-bus, because this code:
```python
BUS_BASE = 'net.tenshu.Terminator2'
BUS_PATH = '/net/tenshu/Terminator2'
try:
    # Try and include the X11 display name in the dbus bus name
    DISPLAY = Gdk.get_display().partition('.')[0]
    # In Python 3, hash() uses a different seed on each run, so use hashlib
    DISPLAY = hashlib.md5(DISPLAY.encode('utf-8')).hexdigest()
    BUS_NAME = '%s%s' % (BUS_BASE, DISPLAY)
except:
    BUS_NAME = BUS_BASE
```
`snapcraft.yaml` require constant name for bus name. Bad app design?

file: `snapcraft.yaml`:
```yaml
name: terminator
version: "1.92"
summary: Terminator is an open-source terminal emulator.
description: |
  Multiple terminals in one window.
  The goal of this project is to produce a useful tool for arranging terminals.
  It is inspired by programs such as gnome-multi-term, quadkonsole,
  etc. in that the main focus is arranging terminals in grids
  (tabs is the most common default method, which Terminator also supports).

#icon: https://raw.githubusercontent.com/gnome-terminator/terminator/master/data/icons/hicolor/scalable/apps/terminator.svg
grade: stable # Use stable or devel
confinement: classic
license: ISC
base: core20

architectures:
  #- build-on: [amd64, armhf]   # Need fix for multipass.
  - build-on: [amd64]
    run-on: [amd64]

apps:
  terminator:
    command: bin/terminator
    desktop: share/applications/terminator.desktop
    environment:
      PYTHONPATH: $SNAP/usr/lib/python3/dist-packages:$SNAP/lib/python3.8/site-packages

parts:
  terminator:
    plugin: python
    source: https://github.com/gnome-terminator/terminator/releases/download/v1.92/terminator-1.92.tar.gz

    build-packages:
      - python3-gi
      - python3-gi-cairo
      - python3-psutil
      - python3-configobj
[...]

Issue #347, opened 2020-12-31
Terminator flatpak-spawn integration (HostCommand)


JayDoubleu:
Flatpak is using the "flatpak-spawn" command to escape from the sandbox and run commands on the host.

Tilix has implemented the below so it automatically spawns shell on the host instead of the container
https://github.com/gnunn1/tilix/blob/787e5335ecef9cada035170eeaedb301f567127d/source/gx/tilix/terminal/terminal.d 

Would it be possible to look at the implementation in terminator ? 
Here is an example of python running flatpak's "HostCommand" 

https://stackoverflow.com/questions/39494813/pass-file-descriptor-via-a-dbus-function-call-from-python-aka-call-flatpaks-ho

Currently I can launch terminator from flatpak but I have to run `flatpak-spawn --host zsh` from within terminator and this is not persistent across tabs.

Alternatively, if there would be a way to pass default run command to terminator as an argument for example:
`--default-shell="flatpak-spawn --host $SHELL"` which would apply to all tabs overwriting the config.

mattrose:
Where did you get a flatpak build of Terminator from?  It's been on my todo list for a while but I've never used flatpak, and I haven't had time to do the research to build up a flatpak build yet.  Did Redhat build one up?  If we can get access to the flatpak build we can suggest a patch for flatpak that will do the right thing for the flatpak package.

The best thing I can think I can think of is to add a profile, or change the default profile to have `flatpak-spawn --host $SHELL` to be the default command that runs on terminal launch.

JayDoubleu:
No, redhat didn't yet build one.
I've had a go at it and it looks promising so far. Have a look here: https://github.com/flatpak-rpm-builds/com.tenshu.terminator

I know that the `toolbox` tool on fedora also uses `flatpak-spawn` so it could be potentially used for development without having to build/rebuild the flatpak.

JayDoubleu:
@mattrose this is now on https://src.fedoraproject.org/flatpaks/terminator/tree/master 
Fedora builds flatpaks as OCI containers from RPM so essentially this is mostly ready for implementation and it will pull your latest RPM is it becomes available in fedora repos (with flatpak-spawn fixes).
From there there is a short route to get it into fedora flatpak registry if folks at fedora decide to include it there.

mattrose:
[...]

PR #570, opened 2022-01-21
Add initial flatpak-spawn support

JayDoubleu:
This is adding experimental flatpak support as well as definition.

It builds and works ok however flatpak-spawn doesn't seem to be handling tty properly (see https://github.com/flatpak/flatpak/issues/3697)

It's usable and perhaps could be submitted to flathub beta and wait until flatpak-spawn resolves issues with tty or someone rewrites it. 

Fixes: https://github.com/gnome-terminator/terminator/issues/206

### DIFF
--- a/terminatorlib/util.py
+++ b/terminatorlib/util.py
@@ -407,20 +407,13 @@ def update_config_to_cell_height(filename):
 
 def get_flatpak_args(args, envv, cwd):
     """Contruct args to be executed via flatpak-spawn"""
-    import json
     flatpak_args = None
-    shell = [args[0]]
     env_args = ['--env={}'.format(env) for env in envv]
     flatpak_spawn = [
         "flatpak-spawn", "--host", "--watch-bus", "--forward-fd=1",
         "--forward-fd=2", "--directory={}".format(cwd)
     ]
-    if len(args) == 2:
-        flatpak_args = flatpak_spawn + env_args + shell
-
-    # Support -x, -e, custom commands etc.
-    if len(args) > 2:
-        flatpak_args = flatpak_spawn + env_args + args[1:]
+    flatpak_args = flatpak_spawn + env_args + args
 
     dbg('returned flatpak args:  %s' % flatpak_args)
 
--- a/terminatorlib/util.py
+++ b/terminatorlib/util.py
@@ -413,6 +413,11 @@ def get_flatpak_args(args, envv, cwd):
         "flatpak-spawn", "--host", "--watch-bus", "--forward-fd=1",
         "--forward-fd=2", "--directory={}".format(cwd)
     ]
+    # Detect and remove duplicate shell in args
+    # to work around vte.spawn_sync() requirement.
+    if len(set([args[0], args[1]])) == 1:
+        del args[0]
+
     flatpak_args = flatpak_spawn + env_args + args
 
     dbg('returned flatpak args:  %s' % flatpak_args)
--- a/terminatorlib/util.py
+++ b/terminatorlib/util.py
@@ -43,6 +43,9 @@ DEBUGCLASSES = []
 # list of methods to show debugging for. empty list means show all methods
 DEBUGMETHODS = []
 
+def is_flatpak():
+    return os.path.exists("/.flatpak-info")
+
 def dbg(log = ""):
     """Print a message if debugging is enabled"""
     if DEBUG:
@@ -144,6 +147,13 @@ def path_lookup(command):
 
 def shell_lookup():
     """Find an appropriate shell for the user"""
+    if is_flatpak():
+        getent = subprocess.check_output([
+            'flatpak-spawn', '--host', 'getent', 'passwd',
+            pwd.getpwuid(os.getuid())[0]
+        ]).decode(encoding='UTF-8').rstrip('\n')
+        shell = getent.split(':')[6]
+        return shell
     try:
         usershell = pwd.getpwuid(os.getuid())[6]
     except KeyError:
@@ -394,3 +404,24 @@ def update_config_to_cell_height(filename):
     except Exception as ex:
         err('Unable to open ‘%s’ for reading and/or writting.\n(%s)'
                 % (filename, ex))
+
+def get_flatpak_args(args, envv, cwd):
+    """Contruct args to be executed via flatpak-spawn"""
+    import json
+    flatpak_args = None
+    shell = [args[0]]
+    env_args = ['--env={}'.format(env) for env in envv]
+    flatpak_spawn = [
+        "flatpak-spawn", "--host", "--watch-bus", "--forward-fd=1",
+        "--forward-fd=2", "--directory={}".format(cwd)
+    ]
+    if len(args) == 2:
+        flatpak_args = flatpak_spawn + env_args + shell
+
+    # Support -x, -e, custom commands etc.
+    if len(args) > 2:
+        flatpak_args = flatpak_spawn + env_args + args[1:]
+
+    dbg('returned flatpak args:  %s' % flatpak_args)
+
+    return flatpak_args
--- a/terminatorlib/terminal.py
+++ b/terminatorlib/terminal.py
@@ -1505,14 +1505,37 @@ class Terminal(Gtk.VBox):
 
         dbg('Forking shell: "%s" with args: %s' % (shell, args))
         args.insert(0, shell)
-        result,  self.pid = self.vte.spawn_sync(Vte.PtyFlags.DEFAULT,
-                                                self.cwd,
-                                                args,
-                                                envv,
-                                                GLib.SpawnFlags.FILE_AND_ARGV_ZERO,
-                                                None,
-                                                None,
-                                                None)
+
+        if util.is_flatpak():
+            dbg('Flatpak detected')
+            args = util.get_flatpak_args(args, envv, self.cwd)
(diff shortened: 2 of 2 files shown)


JayDoubleu:
@mattrose Could you please have a look at the latest changes. I believe this should make it a stable experience when running in flatpak. It also fixes the #347

mattrose:
I will definitely take a look at this, but It'll take me a bit because I want to give it a good once-over, and it's a lot of code to digest in an area I'm unfamiliar with.  Hopefully before the end of the week.  Fedora has included terminator as an rpm at least for the past 10 years or so. @dmaphy and I have been maintaining it.  I'm not sure how to convert the fedora rpm into a flatpak, but maybe we can work that out as well as putting it up on flathub.

JayDoubleu:
No worries, I can help out with building it against fedora runtime as well.
I already opened a repository here https://src.fedoraproject.org/flatpaks/terminator 
It essentially works by sourcing your RPM so if the changes within terminator source code get approved and make it's way to https://src.fedoraproject.org/rpms/terminator then it should be really easy to get it published to fedora.

Nice thing about fedora flatpaks is that once the flatpak manifest has been written and accepted it just uses your RPM to build itself every time there is update to your RPM so you only have to maintain your RPM repo.

JayDoubleu:
PS. I've added you @mattrose  @dmaphy as admins to https://src.fedoraproject.org/flatpaks/terminator and I've also updated it for F35. AFAIK it only needs these steps to run https://fishsoup.net/misc/fedora-docs-flatpak/flatpak/tutorial/#_building_in_koji which I can't do myself as I'm not in maintainers group.

mattrose:
Thanks for all your work on this.

mattrose:
So I've looked this over and I have 2 questions.  We have tried hard to keep packaging files like .spec files and the Debian directory out of this repository.  The Debian packaging files live in the debian repos, and the fedora packaging files live in the fedora repos.  Is there a more appropriate place for the flatpak files to live, rather than in the terminator repo?

Q 2 is more of a concern about the Profiles->Command Preferences pane.  Can you put some indicator in there that the configuration options are not valid when you're running under flatpak.

Other than that, this looks really good.

JayDoubleu:
Hey, my understanding is that there is no need for Flatpak folder to be present within the repo itself at all. It's there only for testing or showcasing it working with flatpak-builder ( flathub native type of flatpak, not OCI )

Fedora flatpaks manage flatpaks via flatpaks/terminator on src.fedoraproject.org (or other repo can be created for this purpose. Fedora runtime flatpaks are OCI type thus use different tooling for building as well as use different manifest format )

If you want to publish to flathub that can be done separately by following their instructions (mainly the flatpak manifest is required for such submission, manifest from this PR could be used for this purpose)


> Profiles->Command Preferences pane. Can you put some indicator in there that the configuration options are not valid when you're running under flatpak.

Would you be able to specify which options are no longer valid ? I was under impressed that functionality was working in latest commit (both running default shell as well as custom commands) or you're referring to things like `-l` flags ?

JayDoubleu:
@mattrose I've rebased and left out all the flatpak/flathub files out of this PR.

It now has only a single commit containing changes required for terminator to work within flatpaks in general. If this gets into a release then building out fedora OCI flatpak should just work.

I've moved the flatpak-builder manifest to my separate branch here: https://github.com/JayDoubleu/terminator/tree/flatpak_flathub_manifest

terminatorlib/util.py
```diff
@@ -407,20 +407,13 @@ def update_config_to_cell_height(filename):
...
         "flatpak-spawn", "--host", "--watch-bus", "--forward-fd=1",
         "--forward-fd=2", "--directory={}".format(cwd)
     ]
-    if len(args) == 2:
-        flatpak_args = flatpak_spawn + env_args + shell
-
-    # Support -x, -e, custom commands etc.
-    if len(args) > 2:
-        flatpak_args = flatpak_spawn + env_args + args[1:]
+    flatpak_args = flatpak_spawn + env_args + args
```
JayDoubleu, line 416: @mattrose this is simplifying the flatpak args. Now it will only prepend flatpak-spawn on front of existing args. I have this if statement there before to avoid running shell with '-c' as flatpak-spawn could just execute commands without it.

With this change flatpak-spawn will always execute `'/bin/bash', '-l', '-c', 'your custom shell command'` when you specify custom command as well as "run as logon shell"

I'm not sure why terminator args are constructed in a way that you always need to run `$shell -c customcommand` when running terminator with things like `-x` and `-e` or custom command in terminator settings.

Do I understand correctly that this is required when using `self.vte.spawn_sync()` ?

Anyway, with this change it has a very simple job which is to take whatever terminator had in args and add it on top of `flatpak-spawn --host --env... args`

This means that when you have no custom settings specified for custom shell instead of executing `/bin/bash` it will execute `flatpak-spawn --host /bin/bash`

When you enable `login_shell = True`:
Instead of running `/bin/bash -l` it will run `flatpak-spawn --host /bin/bash -l`

When you set custom command in settings to /bin/zsh:
Instead of running `/bin/bash -c /bin/zsh` it will run `flatpak-spawn --host /bin/bash -c /bin/zsh`

And with custom command set as well as login shell set : 
Instead of running `/bin/bash -l -c /bin/zsh` it will run `flatpak-spawn --host /bin/bash -l -c /bin/zsh` 

I believe this should make it in line and working with the current setup with all options working.

Perhaps you could consider switching rest of the code to just use `self.vte.spawn_async()` instead of current `self.vte.spawn_sync()` . I saw somewhere that spawn_sync() is deprecated.

mattrose: Thanks, I'll take another look.  The first pass looked like it was just grabbing the user's shell and passing that to vte.spawn_async(), but it's possible I could have missed something.  I'll take another look now. 


The reason terminator doesn't use spawn_async is because it's a relatively new call that isn't supported by Ubuntu 18.04, which is still supported and relatively popular.   What I should do is fix it so that if spawn_async is available then it'll use that and fall back to spawn_sync if they're using an older vte.

JayDoubleu: > The reason terminator doesn't use spawn_async is because it's a relatively new call that isn't supported by Ubuntu 18.04, which is still supported and relatively popular. What I should do is fix it so that if spawn_async is available then it'll use that and fall back to spawn_sync if they're using an older vte.

I see. That makes sense. I guess this problem will be going away with flatpak.

JayDoubleu: Actually It looks like I still need to tweak it making it "simple" to work with all these extra command breaks the normal behaviour with default config :| 

`Terminal::spawn_child: Forking shell: "/bin/bash" with args: ['flatpak-spawn', '--host', '--watch-bus', '--forward-fd=1', '--forward-fd=2', '--directory=/var/home/jaydoubleu', '--env=TERM=xterm-256color', '--env=COLORTERM=truecolor', '--env=PWD=/var/home/jaydoubleu', '--env=TERMINATOR_UUID=urn:uuid:f84be701-e3d3-4d54-862e-dd9d4371dab9', '--env=TERMINATOR_DBUS_NAME=net.tenshu.Terminator25ef4b219e3b005583550f2b0f9f990c3', '--env=TERMINATOR_DBUS_PATH=/net/tenshu/Terminator2', '/bin/bash', '/bin/bash'] via flatpak-spawn`

While the spawn_sync() was working with args like this ` '/bin/bash', '/bin/bash'` the spawn_async() is not happy about it at all.

Going to look into it further

terminatorlib/util.py
```diff
@@ -413,6 +413,11 @@ def get_flatpak_args(args, envv, cwd):
         "flatpak-spawn", "--host", "--watch-bus", "--forward-fd=1",
         "--forward-fd=2", "--directory={}".format(cwd)
     ]
+    # Detect and remove duplicate shell in args
+    # to work around vte.spawn_sync() requirement.
+    if len(set([args[0], args[1]])) == 1:
+        del args[0]
```
JayDoubleu, line 419: This will remove duplicated first command/shells which is caused by https://github.com/gnome-terminator/terminator/blob/master/terminatorlib/terminal.py#L1484 and it's to my understanding it's no longer needed when working with `spawn_async()`

mattrose:
This looks good, though I can't test the flatpak stuff atm.  Is it safe to be merged in?

JayDoubleu:
> This looks good, though I can't test the flatpak stuff atm. Is it safe to be merged in?

I believe it is safe to be merged and should not affect any users who do not use flatpaks.
I've been using this build in flatpak as my daily driver and I had no issues with it so far so I guess it should be good for flathub.

Once merged and this version makes it's way to fedora I will be able to build and switch to fedora OCI flatpak full time and do some further testing.

I guess in both cases the changes would need to be made outside of this repository if there is any tweaks to be done.

JayDoubleu:
@mattrose  the https://github.com/JayDoubleu/terminator/tree/flatpak_flathub_manifest/experimental/flatpak is in line with this PR + flatpak manifest so in case if you would like to quickly test it on some Fedora VM or something you could use instructions from readme.

JayDoubleu:
hey @mattrose it's been a while.. do you think you'll find some time to look into this ?

mattrose:
Sorry, I've been really swamped lately but I hope to look at it soon.

JayDoubleu:
Thanks @mattrose I've been using it since as flatpak and had zero problems with it so far.
If this makes it into a release and then published to fedora or flathub it would be the first stable terminal emulator working as flatpak. I'm sure there is a lot of appetite for it in the community.

Issue #569, opened 2022-01-21
Language not applied after build


pmraps:
Hello,
I tried to build Terminator after updating the pt translation, hoping to get it working for testing purposes. Instead, I get Terminator all in English.
I tried **/po/genpot.sh**, got this:
***xgettext: erro ao abrir "../terminatorlib/encoding.py" para leitura: Ficheiro ou pasta inexistente
ERROR: xgettext failed to generate PO template file. Please consult error message above if there is any.***

Tried **update_all_po.sh** and got a bunch of lines like this one (presumably one for each language):
***msgmerge: erro ao abrir "terminator.pot" para leitura: Ficheiro ou pasta inexistente***
Portuguese parts meaning something like: ***error opening xxx: file or folder inexistent***

Could someone tell me what is wrong, please?
**uname --all**: *Linux debian-stable 5.10.0-10-686-pae #1 SMP Debian 5.10.84-1 (2021-12-08) i686 GNU/Linux*
**locale -a**: 
*C
C.UTF-8
POSIX
pt_PT.utf8*

Thank you!

Vulcalien:
That's something I missed! Thanks for bringing this up.

It's looking for a file that has been deleted in PR #536.

I'll make a PR soon, but you could apply this patch. I assume you have cloned the repository.
`git apply patch.txt` in the repository. Then, running `genpot.sh` and `update_all_po.sh` should work.

[patch.txt](https://github.com/gnome-terminator/terminator/files/7917177/patch.txt)

pmraps:
Great, everything running, thank you.
One last remark, I had to manually copy the .mo file to /usr/share/locale/pt_PT/LC_MESSAGES. It didn't copy automatically, not even with `sudo python3 setup.py install...`
Not sure if this is intentional.

PR #571, opened 2022-01-22
Fix POTFILES.in + update translation files

Vulcalien:
In PR #536 `encoding.py` was removed, but the translation files were still trying to read that file.
That brings to the errors in #569.

This PR simply removed one line in POTFILES.in and the rest is just output of the two scripts (`genpot.sh` and `update_all_po.sh`)

The translations for `encoding.py` are no longer necessary and the script automatically deleted them. I don't know if there is something to do in transifex.


mattrose:
Thanks again!

PR #576, opened 2022-01-28
Remove duplicated info in some debug messages

Vulcalien:
In the code, we can find code like this:
```python
dbg('ConfigBase::get_item: %s found in globals: %s' %
                    (key, self.global_config[key]))
```

But this gets printed:
```
ConfigBase::get_item: ConfigBase::get_item: focus found in globals: click
```

There is a repetition, because the `dbg` function already adds the name of the function at the beginning (4d216633fca8ca46fc3b51fdf2b382db93e3ade1).
The same should probably have been done to the `err` function.

This change should make the debug logs easier to read.


mattrose:
Thanks for this, this is something that was an irritant to me up until now.

Issue #298, opened 2020-11-21
Enhancement: Stjerm Layout Like Functionality


maximstewart:
Would ya'll consider adding layout bind like that of Stjerm? With Stjerm one can pass a position argument of left, right, top, and bottom as well as width and height arguments. From there, you can bind to a toggle key that shows/hides the terminal. What this allows me to do is something like the follow:

exec stjerm -k f8 -p left -h 800 -w 700 &
exec stjerm -k f9  -p right -h 800 -w 700  &
exec stjerm -k f12 -p top -h 500 -w 700 &
exec stjerm -k f10 -p bottom -h 500 -w 700 &

It bind a terminal to the respective key and set the width and height. From there, it's toggleable and shows on the monitor the mouse is on.

For terminator, I know there is the --geometry argument and I know there is a toggle key setting one can set. I don't see a cli argument to set the keybinding. I know that -p is already used to load a profile. So, a different argument is needed there. But, it would be nice to bind unique key's or combo of keys that show or hide their respective terminals. Additionally, positioning the window automatically based on the monitor the mouse is on would be a great addition.

I've been looking through Terminator's code but am a little unclear on where to focus/start.

Monitor data can be gotten with something like this but modified accordingly:

def getMonitorData(self):
    screen    = self.builder.get_object("Main_Window").get_screen()
    monitors = []
    for m in range(screen.get_n_monitors()):
        monitors.append(screen.get_monitor_geometry(m))

    for monitor in monitors:
            print("{}x{}+{}+{}".format(monitor.width, monitor.height, monitor.x, monitor.y))

    return monitors

To get mouse position, something like the following might work. from there, that data is compared to the monitor data and the window is positioned appropriately on the toggle action. The package python-xlib would be needed.

from Xlib import display

def mousePosition():
    data = display.Display().screen().root.query_pointer()._data
    return data["root_x"], data["root_y"]

mattrose:
There's a lot to unpack here but the first big problem I see is that the Xlib stuff most likely won't work on Wayland, or MacOS (It's not officially supported, but I run terminator on my mac a lot). There are Gtk and Gdk calls that will work across all three, though, I'm pretty sure.  That way we also wouldn't need to import another lib.

maximstewart:
[...]

PR #579, opened 2022-02-09
Logic for improved guake like terminal functionality

maximstewart:
I wanted to add an improved Guake like terminal experience that adds options to lock the window to the top, bottom, left, or right part of any monitor. In addition, I wanted to override the toggle button so that the user could bind multiple terminals to separate toggle keys. This is on par with what Stjerm does which is abandoned at this time. 

Edit:  Fixed a stupid mistake after code cleanup. 

Edit: I found the request I made some time back:
https://github.com/gnome-terminator/terminator/issues/298

Yes, it was from an old account that no longer exists. Anyway, this is intended to address the above closed issue/feature request.

### DIFF
--- a/terminatorlib/window.py
+++ b/terminatorlib/window.py
@@ -77,13 +77,16 @@ class Window(Container, Gtk.Window):
 #        self.set_property('allow-shrink', True)  # FIXME FOR GTK3, or do we need this actually?
         icon_to_apply=''
 
-        self.register_callbacks()
         self.apply_config()
 
         self.title = WindowTitle(self)
         self.title.update()
-        
-        self.preventHide = False
+
+        self.preventHide  = False
+
+        self.display      = Gdk.Display().get_default()
+        self.mouse        = self.display.get_default_seat().get_pointer()
+        self.guake_key    = None
 
         options = self.config.options_get()
         if options:
@@ -92,14 +95,43 @@ class Window(Container, Gtk.Window):
 
             if options.role:
                 self.set_role(options.role)
-            
+
             if options.forcedicon is not None:
                 icon_to_apply = options.forcedicon
 
             if options.geometry:
                 if not self.parse_geometry(options.geometry):
-                    err('Window::__init__: Unable to parse geometry: %s' % 
+                    err('Window::__init__: Unable to parse geometry: %s' %
                             options.geometry)
+
+            if options.guake_key:
+                self.guake_key = options.guake_key
+
+                if options.guake_side and options.guake_width and options.guake_height:
+                    proceed_undecorated = True
+
+                    if options.guake_side in ["top", "bottom", "left", "right"]:
+                        self.guake_side = options.guake_side
+                    else:
+                        proceed_undecorated = False
+                        err('Window::__init__: Unable to parse guake_side: %s' %
+                                options.guake_side)
+
+                    try:
+                        self.guake_width  = int(options.guake_width)
+                        self.guake_height = int(options.guake_height)
+                    except Exception as e:
+                        proceed_undecorated = False
+                        err('Window::__init__: Unable to parse guake_width: %s and/or guake_height: %s' %
+                                options.guake_width, options.guake_height)
+
+                    if proceed_undecorated:
+                        self.set_decorated(False)
+                        self.set_default_size(self.guake_width, self.guake_height)
+                    else:
+                        self.guake_key = None
+
+        self.register_callbacks()
 
         self.apply_icon(icon_to_apply)
         self.pending_set_rough_geometry_hint = False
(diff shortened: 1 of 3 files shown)


mattrose:
I really like this PR, and I'll look at it as soon as I can, but I'm kinda busy right now, so I'm not sure when I'll be able to look at it, but hopefully within the next week or so.

Thanks for contributing!

maximstewart:
I did find another weird bug where terminator itself couldn't seem to launch a new instance with the guake width and height arguments because it itself passed a string instead of int type. I changed it to default to string and then do a conversion to int when pulling the guake width and height. Anyway, I believe I've resolved all known issues and pushed a clean and rebased commit.

maximstewart:
Just wanted to update that everything has been stable for me using this "patch" or feature. I've not found any other bugs or issues hat needed my addressing that I haven't already done. Obviously this is just on my system but I don't believe it will be different elsewhere.

mattrose:
Sorry, this is still on my radar, I just haven't had time to look at it yet.  I hope tomorrow I'll have a little bit of free time after doing housework

maximstewart:
@mattrose 

Here is as start command:

terminator --guake-key="F8" --guake-side left --guake-width 700 --guake-height 800

To address your other points:

Editing a config file forces you to open an editor and make changes. Having it this way allows you to quickly spawn an instance as needed. Adding it to the UI would seem to pose a similar issue in my mind. In addition, the main way I use this is through my start script for my system. If I add a config entry then it would seem I'd start  an terminator instance just to start another terminator instance. Though, I could be misunderstanding your intent.

maximstewart:
@mattrose 

I'm not familiar with how terminator config is structured; but, wouldn't setting this in the config mean that each time you run terminator it spawns new instances of the same pinned terminals?  If so, I most certainly want to keep it as cmd args.

mattrose:
I don''t think so.  terminator only uses one process, and just spawns new windows upon repeated calls of terminator (unless you specifically disable dbus).  I will play with it some and get back to you.

maximstewart:
Sounds good, I'll await your findings. But, to ask and address some other points, by adding this to the config variables I would be limited to binding 4 windows, correct? Also, to be frank, I'm unsure of what help text file and man page files to edit. Still working out the project structure a bit.

mattrose:
Hmm, this doesn't seem to work on my F34 desktop with wayland.  Does this require X11 to work?

maximstewart:
@mattrose   Based on the import section at line 29 It is. But, Keybinder could be fixed to work with Wayland at this point. I didn't really look too deep.

maximstewart:
I hope you are doing well. I am just checking in to see if you have done further testing?

maximstewart:
So, sorry for taking so long to get back to this but I added what I think are the needed changes for the man page. I git rebased my commit and added changes to doc/terminator.1. I also did do some slight code cleanup/variable rename as well but nothing major. In addition, I updated some verbiage for terminatorlib/optionparse.py and added an import for RawTextHelpFormatter which allows the argparse add_argument method to format help text as expected when using \\n. (I can remove if not desired but otherwise the example text is long and on the same line as the description.) I found that fix the following SA post:

https://stackoverflow.com/questions/3853722/how-to-insert-newlines-on-argparse-help-text

maximstewart:
Ngl, I almost gave up on this given the misery I had trying to pull new changes, resolve merge conflict, then totally bungle it up, and thereafter try reverting my bungled merge only to revert my work but still pull in the newer changes. I got lucky that github shows the commits even if not tied to a branch.

maximstewart:
🥲

terminatorlib/window.py
```diff
@@ -92,15 +96,44 @@ def __init__(self):
...
             if options.forcedicon is not None:
                 icon_to_apply = options.forcedicon
 
             if options.geometry:
                 if not self.parse_geometry(options.geometry):
-                    err('Window::__init__: Unable to parse geometry: %s' % 
+                    err('Window::__init__: Unable to parse geometry: %s' %
                             options.geometry)
 
+            if options.guake_key:
```
mattrose, line 108: Can we move these command-line options to config file variables?  We can still leave them as command-line options, but merge them into the config file variables with self.config.options_get()

mattrose (changes requested):
The rest of this is ... fine, the whitespace changes are annoying, but I'd really have these be config variables as first-class, with command-line options being variable.

I won't worry about integrating this into the preferences dialog right now, but that's something that should be done down the road.

mattrose:
I'm sorry to be so picky.  I like your changes, and I think they bring out some functionality, but I really think having these as config variables is a much, much better solution for most people.

PR #589, opened 2022-04-09
Zoom on notebook even if there is only one terminal in the tab + keep tab position and label in notebook rotation

Vulcalien:
## The problem
When there are multiple tabs, the notebook top bar is visible.
Now, it is possible to split the terminal and then zoom/maximize, so that the top bar is hidden.
But when there is only one terminal in the notebook tab, it's not possible to do it.

## Solution
I added the signals for the notebook, so that it does zoom/maximize now.

## Current bug
When unzooming a tab **with a single terminal** in it, the tab is moved to the end of the notebook.
I currently have no idea why it does that, but since it can be annoying I think it's best to fix it before merging.

## Second commit
The function I deleted in the second commit (container.py) seems to be unused. Nowhere it's called and other functions do the same thing.

### DIFF
--- a/terminatorlib/container.py
+++ b/terminatorlib/container.py
@@ -134,17 +134,6 @@ class Container(object):
         """Handle a keyboard event requesting a terminal resize"""
         raise NotImplementedError('resizeterm')
 
-    def toggle_zoom(self, widget, fontscale = False):
-        """Toggle the existing zoom state"""
-        try:
-            if self.get_property('term_zoomed'):
-                self.unzoom(widget)
-            else:
-                self.zoom(widget, fontscale)
-        except TypeError:
-            err('Container::toggle_zoom: %s is unable to handle zooming, for \
-            %s' % (self, widget))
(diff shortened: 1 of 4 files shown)


Vulcalien:
Ok I've fixed the bug!

Since I don't think I explained what this PR does nicely, here is what I mean:
zoom-notebook-0

zoom-notebook-1

To zoom, both the key mapping and the right-click GUI menu seem to work fine.

mattrose:
Thank you so much.  Looks good at first glance, I'm gonna pull it down and run it myself to make sure, but at this point you probably know the code as well as anyone else does :smile:

Vulcalien:
I've been using this branch since I made it, and so far I see no bugs in it.
I looked at the code and can confirm that `toggle_zoom` is never used anywhere.

It's not really that useful of a change, more like a "why not" thing.

mattrose:
Looks good, sorry about the wait.

PR #590, opened 2022-04-16
Modification in the /terminatorlib/prefseditor.py file

amaan211:
While editing the preferences in terminator, If you select a particular palette, all of the palette widgets get locked which prevent you to make small changes according to your taste in that particular palette. To do that, you need to select "custom" palette where the widgets get unlocked and you can change the palette. I have changed it so that you can make slight changes then and there and don't have to select "custom" and change it from there. This is more convenient and easier to understand (I was stuck trying to make slight change to the "solarized" palette for a very long time).

### DIFF
--- a/terminatorlib/prefseditor.py
+++ b/terminatorlib/prefseditor.py
@@ -966,10 +966,6 @@ class PrefsEditor:
             if self.palettevalues[key] == active:
                 value = key
 
-        sensitive = value == 'custom'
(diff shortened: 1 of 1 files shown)


mattrose:
Wow, that is pretty nifty.  I'm not too sure how this works so I'm gonna pull down the code and look at it, but thanks in advance!

mattrose:
Thank you!

Issue #606, opened 2022-06-11
Ctrl+Shift+Mouse-Scroll triggers Ctrl+Mouse Scroll


nonZero:
**Describe the bug**
1. While holding Ctrl+Shift+Mouse wheel/Scroll up the font size increases (it shouldn't change, as Shift is pressed)
1. While holding Ctrl+Shift+Mouse wheel/Scroll **DOWN** the font size also **increases** (it shouldn't change, as Shift is pressed)

**To Reproduce**
Use Ctrl+Shift+Mouse wheel/Scroll up/down

**Expected behavior**
Font size should not change.
(It should change when Ctrl+Mouse wheel/Scroll up/down is used, without Shift)

**Desktop (please complete the following information):**
 - Arch Linux
 - Terminator 2.1.1
 - X11

It happens on google chrome too: https://bugs.chromium.org/p/chromium/issues/detail?id=1335654 (but not with firefox or other apps)

Vulcalien:
I can replicate what you say for the shift key, but `Ctrl+Shift+Mouse Down` does not increase the size for me, instead it decreases it as expected.

I think the shift key is being completely ignored. Shouldn't be too hard to fix.

PR #609, opened 2022-06-13
Terminal ctrl+mousewheel: do not try to zoom if shift is pressed

Vulcalien:
Solves #606 (maybe partially)

## The intended feature
To zoom/unzoom the font, it is possible to do `ctrl+mousewheel` in the terminal.
To do fast scrolling, there is `shift+mousewheel`.

## Problem
However, ctrl+shift+mousewheel will zoom/unzoom the font as well, ignoring the shift key.

## Solution
In the code, check if the modifier is *only* `ctrl`, so that the zoom/unzoom is not triggered when doing `ctrl+shift+mousewheel`

## Question
Should we check for things like `ctrl+alt+mousewheel`, `ctrl+super+mousewheel` and so on as well?


mattrose:
I'm so sorry I haven't merged this in, I'm just super-busy right now.   I haven't forgotten about you, I just don't seem to have any time right now.  Thank you so much for. putting in this work.

Vulcalien:
No worries! It's not that urgent anyway, it sure can wait

Issue #608, opened 2022-06-13
should background picture not splited while the terminal is split


meison92:
**Is your feature request related to a problem? Please describe.**
When I split the screen, the background picture is always stretched, which doesn't look very beautiful. 

**Describe the solution you'd like**
Can I make the background picture not splited while the terminal is split

**Describe alternatives you've considered**
one terminator windows use one background img.

**Additional context**
![image](https://user-images.githubusercontent.com/17672199/173341081-8739d35f-72c7-4639-8444-a4d3154da363.png)

Vulcalien:
Hi! That's really a think that should be implemented.

Right now, the image is drawn on the terminal widget, but i think it's totally possible to draw it on notebooks and the window itself.

mattrose:
we could have a background image in the global section, and a different background image in the profile section.  That might look weird in some cases, and that is opening a can of worms layering what would have to be basically transparent vte terminals on top of notebooks or windows with a bg image.  


One other thing we could do is to give a few options on how to handle the image other than just stretch the image, but given the layout presented in this issue, I'm not sure that any way of resizing the image would look acceptable.  As you say, a few non-obvious design choices...

Vulcalien:
I have a bunch of ideas about this. The layering would work perfectly. We just need to write to the notebook/window, without even modifying the way terminals draw.

The image would be visible only through transparent terminals, but that could be set up by the user.
It's the most customizable way I can think of, but since the default for terminals is a solid color, let's remember to put a note like "Remember to set the terminal background to transparent".

Vulcalien:
White we are at it, the code for image stretch is in terminal.py:1126
```python
xratio = float(rect.width) / float(self.background_image.get_width())
yratio = float(rect.height) / float(self.background_image.get_height())
```
Shouldn't be too hard to make a setting to calculate xratio and yratio in different ways.

PR #612, opened 2022-06-21
Window background image

Vulcalien:
Sharing the progress on this issue: #608

So far, these are the changes I've made:
1. add config settings `window_background_type` and `window_background_image`
2. draw a global background image on the window
3. when `Terminator.reconfigure` is called, also apply config to the windows

The feature 3. was added with the first commit. I don't see any reason for the windows not to apply the config without restarting. This makes it so that the global background image changes without restart, but also all the other settings (fullscreen, always on top...).

Still not ready to merge. Missing features:
1. the preferences menu should have a way to set the global background image
2. settings to choose how the image fills the window/terminal (stretch, scale keeping ratio... (?))

Introduced bugs:
1. When a notebook loses focus, the window background becomes a solid color. Doesn't matter if it was transparent or an image.

Also, are the names `window_background_type` and `window_background_image` ok?


Vulcalien:
Here is a config file to help display what this PR does. Also useful for testing.
Change the paths for settings `window_background_image` and `background_image` to whatever image.

[config.txt](https://github.com/gnome-terminator/terminator/files/8949136/config.txt)

Vulcalien:
I updated the Window.background_draw function to match the newer and improved version in Terminal. That version, introduced in #713, increased control over how the terminal's background image is drawn. The same now applies to the window's background image.
This implements one of the missing features listed above (2. _settings to choose how the image fills_...).

The function has practically been copy-pasted from Terminal, so it is a significant code duplication issue. Eventually, if this feature ever gets finished, that should be addressed.

Issue #71, opened 2020-05-11
Scroll to last executed command


mp3guy:
Sometimes when I run commands that output a lot, I find it difficult to scroll to the top of that output and find where in the scroll back I executed the last command. Is there a way of jumping to the scrollback line where the last command was executed?

mattrose:
Unfortunately, the terminal itself has no way of differentiating a prompt from any other text that is on the terminal.  Terminal and iterm2 on OSX have a way of marking prompts, and then being able to jump to the previous mark.  This is a neat feature that I just discovered, and, oddly enough, I don't think any other linux terminal is able to do this.  There may be a technical reason for this, but I may also poke around the iterm2 source code to see if I can code up something similar.

One hack to achieve this would be to do Ctrl-Shift-f and search for your prompt, and this will find your last prompt.

mp3guy:
Even just a feature where it can move the scroll bar back to the line you were at when you last pressed Enter would do the trick. I'd thought about the Ctrl-Shift-f option but was really looking for a single keypress option.

lazyfrosch:
Following prompts will only work when there is some kind of marker. Remembering the scroll position might be tricky...

egmontkob:
See also https://bugzilla.gnome.org/show_bug.cgi?id=767230.

egmontkob:
VTE 0.76 will support this, with the hardcoded Shift+Ctrl+Left/Right keys, using the quasi-standard OSC 133 escape sequences (the prompt is set up in vte.sh to emit these).

egmontkob:
However, Terminator hijacks the Shift+Ctrl+Left/Right (and also Up/Down) keypresses, so this functionality (as well as scrolling by lines) is not available.

Ideally VTE should make these shortcuts configurable, but I'm not sure it's going to happen any time soon. Maybe Terminator should reconsider its own shortcuts. Until then, users can reconfigure Terminator's ones in its settings.

mp3guy:
I just switched to kitty terminal instead

PR #625, opened 2022-07-24
Add "scroll caching" functionality, similar to mark jumping in iTerm2

mp3guy:
Looking for feedback on this addition. First draft so anticipating a lot of questions/changes requested. 

This PR adds this functionality: https://stackoverflow.com/questions/17212643/how-to-scroll-back-the-terminal-to-where-the-last-command-was-given. 

Essentially, every time you press `<Return>`, the current scroll location is cached. Then, by using `<Ctrl>Up` and `<Ctrl>Down` you can navigate back and forward through this location history. This is useful for scrolling back to the beginning of the output of previously run commands. 

As far as I can tell, this functionality cannot be added in a plugin as the integration required is a bit too deep? Happy to be corrected here. 

I have tested this with and without infinite scrollback and also ensured it's robust to the user calling `clear` and `reset` in the terminal. It is also robust to out-of-order cache insertions that may happen e.g. if you use `fzf` to do command selection before running a command. Of course, exotic things like ncurses and other non-linear terminal interfaces will not really make sense with this functionality. 

Awaiting feedback, hopefully this feature is useful to others.


mattrose:
Hopefully fixes #71.  Thanks so much for "scratching your own itch" on this and getting some code out.  I definitely have a few suggestions for improvements, but this is definitely almost there.

mp3guy:
Actually I noticed some weird stuff happening when running things like vim. I'll close this and re-open when I have a better solution. The iterm2 solution of a hidden string at the prompt mark seems like the most robust solution: https://iterm2.com/documentation-shell-integration.html

mattrose:
Yeah, I may have to try and setup terminator to receive OSC sequences itself.  I've been relying on VTE to do that up until now, but I might look at trying to intercept and act on those sequences within terminator itself.

mp3guy:
After poking around I didn't see a way to do it, the VTE is really unforgiving and inflexible. I am almost at a solution that seems robust, it does involve adding a hidden marker to your `PS1` just like iTerm2 however. Just a couple edge cases left I'm trying to work around. It's bizarre that it's 2022 and this isn't a widespread feature. 

There is another solution that is much easier which is leveraging the search previous/next functionality within VTE combined with the hidden marker, this is very small implementation wise and works with all/any terminal applications but has some annoying quirks that make it not quite as nice.

mattrose:
VTE is a very ... opinionated library.  It reflects it's maintainer like that 😄 

So, I have started to figure out a way to read the OSC sequences from any text coming in, I just have to figure out a way to feed the escape sequences back.

I like your idea of using <Ctrl><up> and down to search through marks.  That pretty much mirrors iTerm and Terminal.app.

PR #628, opened 2022-07-29
Fix missing icons when started with Ctrl-Alt-T

MihaiBabiac:
By default, GTK sets the prgname to argv[0], which is then used to choose the correct .desktop file. If the program is run through a symlink, such as through `/etc/alternatives/x-terminal-emulator`, argv[0] is not `terminator` so under Wayland the desktop file is not detected, leading to a missing icon.

Possibly fixes these issues:
https://askubuntu.com/questions/1218884/terminator-does-not-add-to-dock-icon-when-opened-via-ctrlaltt
https://bugs.launchpad.net/terminator/+bug/1726525

### DIFF
--- a/terminator
+++ b/terminator
@@ -33,7 +33,7 @@ try:
     import gi
     gi.require_version('Gtk','3.0')
     # pylint: disable-msg=W0611
-    from gi.repository import Gtk, Gdk
+    from gi.repository import Gtk, Gdk, GLib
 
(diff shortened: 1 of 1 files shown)


francocm:
Tested on a fresh Ubuntu 22.04.1 install and works as expected.

mattrose:
Thanks for tracking this down!  I've been trying to track down the cause for this for a while

PR #632, opened 2022-08-15
this line has an extra ';' symbol

xuezhixin:


mattrose:
don't see the harm in this, thanks for fixing this up

PR #653, opened 2022-10-15
- bug context menu (right click)->layouts->"Layout Name" always selec…

vssdeo:
…ts the last Layout

- fixed so the right layout is picked and loaded


Vulcalien:
Can reproduce. After a really quick test, the change seems to work.

mattrose:
Thanks for the fix @vssdeo and thanks for confirming this @Vulcalien!

vssdeo:
cheers @mattrose glad to be of any help.

PR #655, opened 2022-10-16
Plugin Submission : SaveLastSessionLayout Uses Layout to Auto-Save Last session and CWD

vssdeo:
- added plugin save_last_session_layout.py
- saves the working directory and layout of last session (last window) closed
- layout is saved under name of SaveLastSessionLayout
- can be loaded from context menu (right click) -> layouts -> "SaveLastSessionLayout"
- last working directory is also saved

### DIFF
--- a/terminatorlib/container.py
+++ b/terminatorlib/container.py
@@ -245,7 +245,7 @@ the tab will also close all terminals within it.')
 
         return(terminals)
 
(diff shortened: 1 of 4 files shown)


mattrose:
I've been unable to reproduce all of this with my checked out code, but I really like this and am merging it anyways.  I'll poke at it further once merged

PR #658, opened 2022-10-23
Update colour to color to be consistent

wishdev:
I am attempting to make some changes to the way palettes are handled via the config screen and the switching between colour and color is a little confusing. This simply makes it fairly consistent across the board.

I did not change the Changelogs nor the en_GB translation.


nautics889 (approved):
Looks like all occurrences have been replaced.
>I did not change the Changelogs nor the en_GB translation.

I don't think they have to be changed (especially en_GB :) ). 
___
Thanks!

PR #659, opened 2022-10-27
Update terminal.py

flaviosteimacher:
Fix error insert terminal number


mattrose:
Thanks for tracking this down.  I thought I had patched this years ago, but I do keep seeing it pop up

PR #661, opened 2022-10-30
[bug 654] - Plugin Submission : SaveLastSessionLayout Uses Layout to …

vssdeo:
I had not pushed the changes for pre-close-event as I had been waiting for the team. This should work now.

…Auto-Save Last session and CWD on Terminal Window Close

- changes for the pre-close-term event in the plugin
- submitting as I had first thought of using current event close-term, but later thought of a consistent solution


mattrose:
Merging blindly.

mattrose:
Blind merge worth it.  It works much better now.

vssdeo:
> Merging blindly.

working with this team would be fun :)

Issue #662, opened 2022-10-31
[Feature Request] - In the Context Menu(Right-Click) show keyboard shortcuts / accelarators 


vssdeo:
Sometimes remembering short-cuts is handy though in case user forgets, one have to go and check keybindings. It would be handy to display the key shortcuts beside the menu item.

Currently I don't see the short-cut keys visible while I right-click the context menu. This is a standard feature in all other terminal emulators. 

- I have a patch ready in case this seems suggestion seems ok and accepted.
- The patch takes the action from config and updates the menu dynamically

Note: I think there is a bug or discrepancy where for the action edit window title config.py says: 'edit_window_title': '<Control><Alt>w where as in code: terminal_popup_menu.py
 item = Gtk.ImageMenuItem.new_with_mnemonic(_('Set W_indow Title'))

Character i is selected as mnemonic.

- In the patch I have fixed it to Character w as per the config.py

vssdeo:
attachment after applying the patch

![image0](https://user-images.githubusercontent.com/19357612/198946188-b9d2cf72-82ca-4806-a1a8-c205d3b8b15b.jpeg)

mattrose:
Ooo, I like it.  I'll pull it in and check it out soon

vssdeo:
The problem in displaying is happening as the the config the key for eg Control is being saved as <Primary> and the short cut alphabetic character is being picked when creating the menu item and not being overwritten from config. Will check further and fix since this <primary> thing is new to me.

PR #663, opened 2022-10-31
[bug 662] [Feature Request] - In the Context Menu(Right-Click) show k…

vssdeo:
…eyboard shortcuts / accelarators #662

- added short cuts / accelarators to right click context menu
- menu is updated from config so changes in preferences keybindings appear dynamically
- Note: I think there is a bug or discrepancy where for the action edit window title config.py says: 'edit_window_title': 'w where as in code: terminal_popup_menu.py item = Gtk.ImageMenuItem.new_with_mnemonic(_('Set W_indow Title')) Character i is selected as mnemonic.
- fixed the above discrepancy

### DIFF
--- a/terminatorlib/terminal_popup_menu.py
+++ b/terminatorlib/terminal_popup_menu.py
@@ -3,7 +3,7 @@
 """terminal_popup_menu.py - classes necessary to provide a terminal context 
 menu"""
 
-from gi.repository import Gtk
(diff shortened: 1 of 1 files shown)


vssdeo:
Sharpening my own tool ! when nothing else to do ! a selfish act :) Not without a great support from your side ! You are most welcome !

Issue #559, opened 2021-12-20
Add menu autocomplete


entrity:
**Is your feature request related to a problem? Please describe.**
I'm frustrated when I want to perform an action but can't remember how to do it (whether it be a keybinding or an item under some menu).

**Describe the solution you'd like**
I'd like to be able to type what I want into a search field and have that field populate a list of menu items or keybinding which fuzzy-match my search.

The kind of functionality I would like is what sublime text offers when I press ctrl+shift+p or what [Windows terminal](https://github.com/microsoft/terminal) offers (again, when I press ctrl+shift+p).

**Describe alternatives you've considered**
I've not thought of anything short of memorizing keybindings better.

**Additional context**
Here's a screenshot of what sublime text provides:
![sublime-20 11-15-34](https://user-images.githubusercontent.com/950824/146820722-cd82b2e8-ecc6-4584-b292-3858c421a5d1.png)

mattrose:
This is a fantastic idea, and I was thinking of adding something like this as well. PRs are welcome.

PR #665, opened 2022-11-01
[bug 559] Add menu autocomplete #559

vssdeo:
- added a filter to the preferences menu for easy access to shortcuts
- may require some tweak, on search term size and update
Note: when edited in prefs it seems the Short Cut Labels in (right click) Context Menu are not updated will fix that in that bug Issue #662

### DIFF
--- a/terminatorlib/preferences.glade
+++ b/terminatorlib/preferences.glade
@@ -3720,6 +3720,21 @@
               </packing>
             </child>
             <child>
+              <object class="GtkVBox" id="vbox124">
+               <child>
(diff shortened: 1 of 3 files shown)


vssdeo:
Looking into tests.

mattrose:
Thanks for all your work on this.  I'm gonna see if I can open up the preferences window directly to the keybindings page, or create a new window on Ctrl-Shift-P that would have the same contents.  This is actually the hard part though, so thank you!

vssdeo:
@mattrose you are welcome

PR #666, opened 2022-11-01
[bug 662] [Feature Request] - In the Context Menu(Right-Click) show k…

vssdeo:
…eyboard shortcuts / accelarators #662

-fixed: The problem in displaying is happening as the the config the key for eg Control is being saved as <Primary> and the short cut alphabetic character is being picked when creating the menu item and not being overwritten from config
-Note: the short cut for function keys are not being displayed in context-menu


mattrose:
Thanks for this!

vssdeo:
cheers !

Issue #660, opened 2022-10-27
terminator: error: unrecognized arguments


planet36:
**Describe the bug**
When the `-x` option is used and the EXECUTE parameter has options, terminator tries to parse them as its own options.

**To Reproduce**
`terminator -u -g /dev/null -x bash -l`

**Expected behavior**
Terminator starts.

**Desktop (please complete the following information):**
 - Arch Linux
 - Display Technology: X11

**Additional context**
The help message says:
>  -x EXECUTE, --execute EXECUTE
>                        Use the rest of the command line as a command to execute inside the terminal, and its arguments

planet36:
`terminator --version`
> terminator 2.1.2

ncmprhnsbl:
just saying: 
`terminator -u -g /dev/null -x 'bash -l'`
seems to avoid this..

planet36:
> just saying: `terminator -u -g /dev/null -x 'bash -l'` seems to avoid this..

In a previous version that wasn't necessary.

Vulcalien:
About 1 year ago I made a major rewrite of the option parser in #506.

But how do we distinguish between -l being an option for terminator or for bash?
I've seen some program use a `;` to close the command.

I suspect this could be fixed with a one-line change.

mattrose:
The way it's documented is that anything after --execute is the command you're supposed to run, so assume --execute is the last terminator arg.

shawn-ogg:
I ran into this issue because I wanted to use terminator together with ranger integrated in firefox :smile:

mattrose:
Thanks for your code @shawn-ogg.  I merged it into master.

ManuelSchneid3r:
This is still an issue in the recent versions.

AdamPS:
> This is still an issue in the recent versions.

This is expected. The patch was merged in Nov 2022, and the most recent release was Oct 2022. It would be great to have a new release😃.

PR #678, opened 2022-11-18
Fix argument handling of the --execute flag

shawn-ogg:
--execute behaves now like in the docs:
	Runs  the rest of the command line instead of your default shell
        or profile specified command.

This fixes it's proper usage with tools like gio or xdg-open

#660


mattrose:
I love it, thank you!

Issue #681, opened 2022-11-19
Plugin Submission + Generic Plugin Utility Functions & KeyBinding Feature:   Mouseless / Mousefree / Keyboard URL opening or yanking 


vssdeo:
Opening a new ticket to submit code, earlier #450 Mouseless / Free URL opening or yanking may be this one also #675

I am adding a new ticket as this contains 3 things. 

### Features **See Screenshots**

1) Mouse less/free opening / yanking feature
2) **Generic** plugin_util  file to assist this and other plugins with key-press **Helper** function 
3) **Generic** adding **Key Binding Help Functions** in above (2) to config and preferences->keybinding to have a consistent behavior

Code Example: 
terminatorlib/plugin.py
terminatorlib/plugins/mousefree_url_handler.py 

### Testing
**Test Behavior** 

1) Launch the ./terminator & Go to Preferences->Plugins. Enable **MouseFreeURLHandler**
2) Default keys are 
<Alt>j   find next url
<Alt>k  find prev url
Esc      clear last search
<Alt>Return   open the selected URL

Normal Default <Shift><Ctrl>v  paste last selected URL

3. Type https://yahoo.com and press enter on terminal, it will throw error and URL will be on screen
    Press enter , enter 
    Type https://google.com and press enter to have another URL on screen

4. Press <Alt>j to select next URL on screen, <Alt>k to select prev URL on screen, Press <Esc> to clear and <Alt><Enter> to open URL

### Additional Testing:
You can bind keys and key actions from plugins and they will show in Preferences->keybindings
- Go  there and change the key bindings, lets say you choose <Alt>g for next, <Alt>h for prev following will be in your config:-
                [keybindings]
                plugin_url_find_next = <Alt>g
                plugin_url_find_prev = <Alt>h
- Check config 
- Try new bindings 

### Known Issues
- works on new opened tabs but not on newly launched windows, signal copying issue, we were discussing and I had given a patch in  #629
-  not an issue, but had to remove cyclic dependency between config.py and plugin.py

### Improvements
- Can URL selection be directly from vte using Key Press than taking terminal text and then regex to make more efficient
- More Testing on Key Bindings as code touches, Config, Prefs Editor etc.

![terminal_selected_url](https://user-images.githubusercontent.com/19357612/202856719-77c58b6f-a2fd-4e9c-8069-a8ae092dd3bc.png)

[...]

PR #682, opened 2022-11-19
681 plugin submission plugin utility key binding mouseless keyboard url open

vssdeo:
Please Read: https://github.com/gnome-terminator/terminator/issues/681 for the changes.


vssdeo:
Yes it’s already these for testing and comments / changes

vssdeo:
While you are testing, do try, may be playing with two plugins in parallel may throw out more errors or conditions etc. I haven’t included this in PR and this sits in my branch.

https://github.com/vssdeo/terminator/tree/new-selecting-text-with-keyboard-shift-up-shift-down

mattrose:
I had to throw the signal patch into the branch, and make a few more adjustments to terminatorlib/terminator.py to get it to work properly on any subsequent window but the 1st one. (I had to move the copy_other_signals method call from create_layout to new_window.)

I still would rather have the window get the signals from the plugin.  Getting it from the first window feels ... random somehow.  I tried messing with the plugin.py code to get it to load the signals from there, but have so far been unsuccessful.

vssdeo:
I think I tried with re-loading plugins, can’t recall. Will try once back. 

There are use cases for both methods, like fork copying all structures for parent or the other where a new instance should be a new fresh start as per configs. Let me get back home in a day or two and try.

vssdeo:
Hi @mattrose, I had done this to play around with reloading before, this was not final though, just to experiment. Calling the loaded module func was calling plugin ` __init__` again and worked on new window. I still need to check further with import mechanism and any side effects.

```
diff --git a/terminatorlib/plugin.py b/terminatorlib/plugin.py
index 8ab8ddf8..9b900e1e 100644
--- a/terminatorlib/plugin.py
+++ b/terminatorlib/plugin.py
@@ -73,7 +73,7 @@ class PluginRegistry(borg.Borg):
         """Load all plugins present in the plugins/ directory in our module"""
         if self.done:
             dbg('Already loaded')
-            return
+            #return
 
         config = Config()
 
@@ -93,9 +93,11 @@ class PluginRegistry(borg.Borg):
                     try:
                         module = __import__(plugin[:-3], None, None, [''])
                         for item in getattr(module, 'AVAILABLE'):
+                            func = getattr(module, item)
                             if item not in list(self.available_plugins.keys()):
-                                func = getattr(module, item)
                                 self.available_plugins[item] = func
+                            else:
+                                func()
 
                             if item not in config['enabled_plugins']:
                                 dbg('plugin %s not enabled, skipping' % item)

```

vssdeo:
great ! so will you be making the change @mattrose along with your other changes ?

mattrose:
Sure, I can pull this together.   I'll work something up as soon as I can.  Do you mind if I rebase the commit and split it up so that the plugin itself is separate from the supporting code for it?  I might even do the plugin as a separate PR.

Once that's done you should also be able to pull in the other plugin that you've been working on as a separate PR as well.

vssdeo:
Sure @mattrose you are the boss, makes sense.

You mean the other mouse free selection :) yes hope other folks find it of any use. 

Anyways so let me know if anything from my side is needed !

vssdeo:
closing this as this code is decoupled into diff components

Issue #680, opened 2022-11-19
Open up keybindings page on keypress


mattrose:
In #665 @vssdeo implemented a search feature for keybindings, and now I want to open up the preferences window directly to the keybindings page, or create a new window on Ctrl-Shift-P that would have the same contents, for easier access.

vssdeo:
seems the key combo is already taken 'go_prev'          : '<Shift><Control>p'

How about : '<Shift><Control>k', I will make the change for now. 

But adding key seq to a long list of keypress function breaks my heart, If the plugin util keybind makes it to the main branch, I will take this out and may be have a shortcut plugin to handle these.

- Also selecting of Keybindings page is being done for this specific case Page = 3
- A notebook tab number and title map would remove this. Which can be configurable, later.

PR #686, opened 2022-11-23
[bug 680] Open up keybindings page on keypress #680

vssdeo:
- implemented the short-cut
- seems the key combo is already taken 'go_prev' : 'p'
- How about : 'k', I will make the change for now.
- But adding key seq to a long list of keypress function breaks my heart, If the plugin util keybind makes it to the main branch, I will take this out and may be have a shortcut plugin to handle these.
- Also selecting of Keybindings page is being done for this specific case Page = 3
- A notebook tab number and title map would remove this. Which can be configurable, later.

### DIFF
--- a/terminatorlib/config.py
+++ b/terminatorlib/config.py
@@ -202,6 +202,7 @@ DEFAULTS = {
             'next_profile'     : '',
             'previous_profile' : '', 
             'preferences'      : '',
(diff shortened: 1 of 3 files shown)


mattrose:
Thank you!

vssdeo:
Welcome !

Issue #685, opened 2022-11-22
Broadcast to temporary groups


nanborges:
**Is your feature request related to a problem? Please describe.**
When we want to broadcast a command to multiples terminals, we have to create a group and add the terminals, and the send the commands. And sometimes we only want to send a few command, we don't need the group.

**Describe the solution you'd like**
I would like a feature in which I can select terminals, like using "CTRL+Click", and automatically creates a temporary group, so  can send commands to the terminals and after I finish, I can undo de temporary group

I don't know if Terminator already has a feature like this, me and a couple other students are trying to develop something about that.

mattrose:
@nanborges were you working on your own PR for this issue?

mattrose:
Awesome I'll just merge his PR then

PR #691, opened 2022-12-06
Ctrl+Click on group button automatically creates groups whenever needed

nicbn:
When pressing Ctrl+Click on the group button at the top of a terminal, if both the current terminal and the target terminal are not assigned a group:

- Before this change, nothing would happen.
- After this change, a new group with a randomized name will be created, without prompting the user for the group name.

Suggestion made at issue #685, with the goal of improving ergonomics.

Demonstration:

### DIFF
--- a/terminatorlib/terminal.py
+++ b/terminatorlib/terminal.py
@@ -879,7 +879,13 @@ class Terminal(Gtk.VBox):
                 focused=self.get_toplevel().get_focussed_terminal()
                 if focused in targets: targets.remove(focused)
                 if self != focused:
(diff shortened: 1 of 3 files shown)


mattrose:
I like it! I'll merge it down and take a look at it soon.

mattrose:
@nanborges opened #685 with an eye to their own solution, so I've asked them if they're working on something on matrix , so if I don't hear back from them in a few weeks I'll merge this in.

mattrose:
Sorry for the wait on this one

PR #709, opened 2023-02-03
Better distinguishing of inactive windows from the active one, by changing the background brightness

KKoovalsky:
Hello!

I normally don't use title bar, thus differentiation between active and inactive windows, when there is not much in the foreground, is hard for me.

To better explain this, check out this GIF - the problem:

When there is not much data on the terminals, then it's hard to distinguish in which terminal I'm currently writing.

With this PR, one can change the background brightness of the Inactive window:

I have introduced a setting which allows changing the background brightness of the inactive window:

This PR introduces changes to the background color, and the implementation is a simple copy-pasted solution, scrapped from the `inactive_color_offset`, which controls the brightness of the font for an inactive window.

At this stage, it is a draft, because I have few questions:
1. Do you find it good?
2. Should I update the translations as well, since a new text within the "Preferences" window is introduced? Even, when I generate the translation script from the upstream/master, I get a non-null git diff. Therefore, I am "afraid" of putting the updated to the PR.

### DIFF
--- a/terminatorlib/config.py
+++ b/terminatorlib/config.py
@@ -103,7 +103,7 @@ DEFAULTS = {
             'custom_url_handler'    : '',
             'disable_real_transparency' : False,
             'inactive_color_offset': 0.8,
-            'inactive_bg_color_offset': 0.8,
+            'inactive_bg_color_offset': 1.0,
             'enabled_plugins'       : ['LaunchpadBugURLHandler',
                                        'LaunchpadCodeURLHandler',
                                        'APTURLHandler'],
--- a/terminatorlib/config.py
+++ b/terminatorlib/config.py
@@ -103,6 +103,7 @@ DEFAULTS = {
             'custom_url_handler'    : '',
             'disable_real_transparency' : False,
             'inactive_color_offset': 0.8,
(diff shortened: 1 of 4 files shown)


Vulcalien (approved):
It works as intended, great!
It's a feature that should obviously be present.

The only thing to consider is if the default should be 80% (as it is in the PR) or 100% (current behavior).

In preferences.glade some widgets got shuffled by the tool. That might create some conflicts with other PRs.

KKoovalsky:
Cool, that it can be introduced!

I have changed the default value to 100%.

Regarding the `preferences.glade`, what can I do to fix it?

mattrose:
It's been a bit since I touched that code, but how does it work with transparent or image backgrounds?  Does it just set the bgcolor to black?  Can you try this out?

mattrose:
I gave it a shot with both transparent and image backgrounds and it doesn't seem to work.

This is ... not entirely unexpected, and honestly, it's probably fine, as the unfocussed code is going to be re-written at some point in the GTK4 process anyways.  I have some ideas on how to fix this, but that way lies madness.

mattrose:
Anyway, that said, I'm fine with this if you are.

PR #716, opened 2023-03-01
Remove hack to ensure that focus is set.

### LINKED ISSUE #714: "Don't ask, don't tell" focus hack causes gnome-shell to freeze for seconds / Xwayland to crash
Lately I have been having these weird issues where gnome-shell freezes for second and often Xwayland crashes. I have been working with the mutter + Xwayland devs to get to the bottom of this.

Let me copy and paste the summary which I added to a related (workaround) mutter pull-request for this:

I have been seeing mutter freezing for seconds / Xwayland crashes / mutter crashes when starting terminator for a while now.

After collecting backtraces I emailed @carlosg and @ofourdan to see if we could make sense of this.

The Xwayland logs showed a crash due to running out of libwayland buffer space and the crashes actually go away with: https://gitlab.freedesktop.org/wayland/wayland/-/merge_requests/188 which makes libwayland dynamically grow the buffer.

The mutter backtrace showed mutter crashing inside meta-x11-display: update_cursor_theme(), it seems that if Xwayland crashes at the wrong moment it manages to take mutter down with it. These too appear to be resolved by the libwayland change to dynamically grow the buffer.

If Xwayland no longer crashes then things just freeze for a couple of seconds when starting terminator and after that everything works normally again.

Commenting out the g_signal_connect_object("cursor-updated", update_cursor_theme) call can be used as a workaround for the freezes.

@ofourdan has managed to reproduce this by using my ~/.config/terminator/config file. I will attach that here. For me this reproduces by just starting terminator with this config in a gnome-shell which has already had at least one other app opened. I'm on a fully up2date Fedora 37 install.

Further debugging has turned out that this is caused by repeated `gdk_window_focus()` API calls. Which lead me to: https://github.com/gnome-terminator/terminator/blob/403bf540d08920971924e522bd8395485d895581/terminatorlib/terminator.py#L349 and commenting that out indeed fixes the freezes caused by starting terminator.

It looks like to me like this workaround is causing more harm then whatever focus issue it is trying to fix., so maybe remove it?

You can reach out to the gnome-shell devs in: https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/2849 Where a workaround for this terminator mis-behavior is being discussed. I'm sure they would be happy to discuss a better way to deal with whatever focus issue this is trying to fix (or at least fix it under gnome-shell).

mattrose:
set_focus() should do the trick according to the docs, so let's just use that, and not go to extraordinary measures.

fixes #714 hopefully

### DIFF
--- a/terminatorlib/terminator.py
+++ b/terminatorlib/terminator.py
@@ -309,6 +309,7 @@ class Terminator(Borg):
         """Layout operations have finished, record that fact"""
         self.doing_layout = False
         maker = Factory()
+        t = 0
 
         window_last_active_term_mapping = {}
         for window in self.windows:
@@ -350,7 +351,7 @@ class Terminator(Borg):
         # last_active_window is focussed. 
         if self.last_active_window:
             window = self.find_window_by_uuid(self.last_active_window.urn)
-            window.present_with_time(0)
+            window.present_with_time(t)
             window.set_focus()
         self.prelayout_windows = None
 
--- a/terminatorlib/terminator.py
+++ b/terminatorlib/terminator.py
@@ -346,22 +346,11 @@ class Terminator(Borg):
                 t = 0
             window.get_window().focus(t)
 
-        # Awful workaround to be sure that the last focused window is actually the one focused.
-        # Don't ask, don't tell policy on this. Even this is not 100%
+        # Going by the docs, this should be all that's needed to ensure that the
+        # last_active_window is focussed. 
         if self.last_active_window:
             window = self.find_window_by_uuid(self.last_active_window.urn)
-            count = 0
-            while count < 1000 and Gtk.events_pending():
-                count += 1
-                Gtk.main_iteration_do(False)
-                window.show()
-                window.grab_focus()
-                try:
-                    t = GdkX11.x11_get_server_time(window.get_window())
-                except (NameError,TypeError, AttributeError):
-                    t = 0
-                window.get_window().focus(t)
-
+            window.set_focus()
         self.prelayout_windows = None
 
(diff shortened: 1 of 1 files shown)


Vulcalien:
I lost focus on start after this. I'm on X11, Gnome, Debian 12. I'll try to fix it.

mattrose:
> Removing the line `window.set_focus()`, it then works fine for me.

Oo, can you send in a PR with that and I'll merge it?

mattrose:
Days like this I love GTK.  This function does the exact opposite of what the function says it does.  😛

Vulcalien:
I think the "restore focus" feature isn't working altogether.
That part of the code is meant to focus the last active terminal in a layout, right?

If that's so, then I believe that instead of focusing the `last active window`, we should focus the last window's `last active terminal'.

mattrose:
@Vulcalien do you have a config that can reproduce the problem, or is it now fixed with the final change?

Vulcalien:
[config.txt](https://github.com/gnome-terminator/terminator/files/11068746/config.txt)

Here there are two windows, each with one terminal.
From what I read in the config file, the terminal named "This terminal should be focused" should be.. well, focused. Instead, the focus goes to the unnamed window (at least in my machine).

***

[config.txt](https://github.com/gnome-terminator/terminator/files/11068776/config.txt)

For some reason, having a notebook influences how the focus is chosen. In this case, it works as it should.
In other tests ([config.txt](https://github.com/gnome-terminator/terminator/files/11068779/config.txt)), it doesn't.

***

What I think it's happening, is that the window that is written as last in the config file is the one that grabs the focus, every time.
I've tried looking for the cause, but found nothing.

PR #726, opened 2023-03-11
Do not remove focus from the last_active_window

Vulcalien:
This caused all the new windows to be unfocused, forcing the user to manually focus one.

From the [docs](https://lazka.github.io/pgi-docs/Gtk-3.0/classes/Window.html#Gtk.Window.set_focus):
> If `focus` is `None`, unsets the focus widget for this window.

On my machine, the last_active_window feature seems to be broken.
I've spent some time trying to figure out why, but found nothing.

Most of the `layout_done` function appears to do nothing useful, but that might be different on other Desktop Environments.

It's an old problem, with commits from 10 years ago trying to fix it. And it's a mess.

### DIFF
--- a/terminatorlib/terminator.py
+++ b/terminatorlib/terminator.py
@@ -355,7 +355,6 @@ class Terminator(Borg):
         if self.last_active_window:
             window = self.find_window_by_uuid(self.last_active_window.urn)
             window.present_with_time(t)
-            window.set_focus()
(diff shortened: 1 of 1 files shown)


mattrose:
Thanks for doing this. I haven't actually tested it I just merged it from my phone but I looked over the change and it looks good

PR #733, opened 2023-03-30
Rewrite terminator.1 man page in AsciiDoc format

Vulcalien:
`terminator.1` is the man page that opens after typing `man terminator`.

It currently contains a few problems, such as a missing newline, options and their arguments being all in bold instead of bold and italic, keybindings out of place or in the wrong sub-section...
All of that should now be fixed.

The AsciiDoc format seems to me way more friendly than the manpage one.
To generate the manpage file, we just need to run `asciidoctor -b manpage terminator.adoc` after modifying the source.

Maybe it would be a good idea to add a script that runs that command.

***

I've started working on a manpage for remotinator as well. And ideally, `terminator_config` should be rewritten too, but that requires way more work and updates.

### DIFF
--- a/doc/terminator.1
+++ b/doc/terminator.1
@@ -1,286 +1,481 @@
-.TH "TERMINATOR" "1" "Jan 5, 2008" "" ""
+'\" t
+.\"     Title: terminator
+.\"    Author: [see the "AUTHOR(S)" section]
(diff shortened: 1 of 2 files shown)


mattrose:
Documentation Improvements!?  

You're awesome, thanks!

Issue #676, opened 2022-11-17
Title field not seem to be working with JSON profile


tomers:
```
{
  "layout": {
    "tab1": [
      {
        "command": "while true; do reset; docker logs -f kaka; sleep 2; done"
      },
      {
        "command": "while true; do reset; docker logs -f db; sleep 2; done",
        "title": "XXXXXXXXXXXXXXX"
      },
      {
        "command": "bash"
      }
    ]
  }
}
```

```
terminator --config-json ./3-row.json
```
I see the layout properly, including docker logs, but I don't see the title

mattrose:
Merged

PR #736, opened 2023-04-02
Fix for issue #676: title field not seem to be working with JSON profile

rkashinin:


nautics889 (changes requested):
Verified on Ubuntu 22.04, it works correct, the tab has it's title. Thank you for contributing!
___
However there are several points remaining unclear for me.

The function `from_json()` accepts two arguments `layout_name` and `json_name` (default `None`). As i understood the purpose of `json_name` if to distinguish the name of property in layout from the actual key provided in JSON. Can they be different anyway? In your implementation there aren't calls `from_json()` passing `json_name` as an agrument (so `json_name` would equal to `layout_name` all the time). Does `json_name` really matter? 

Also i'm a bit confused about using `__setitem__()` instead of `children[parent + "." + str(order)][layout_name] = layoutjson[json_name]`.

I'd suggest the next implementation:
```
    def build_terminal_layout(self, layoutjson, children, parent, order):
        dbg ('Building a terminal from json: %s' % layoutjson)

        children[parent + "." + str(order)] = {
            'type': 'Terminal',
            'order': order,
            'parent': parent,
            'profile': self.profile_to_use
        }
        for item in ('command', 'title'):
            if item in layoutjson:
                children[parent + "." + str(order)][item] = layoutjson[item]
```
or even like this (a bit prettier as for me):
```
    def build_terminal_layout(self, layoutjson, children, parent, order):
        dbg ('Building a terminal from json: %s' % layoutjson)

        child_conf = {
            'type': 'Terminal',
            'order': order,
            'parent': parent,
            'profile': self.profile_to_use
        }
        for item in ('command', 'title'):
            if item in layoutjson:
                child_conf[item] = layoutjson[item]

        children[".".join([parent, str(order)])] = child_conf
```

(I've checked the code above, it works correct)

Please, let me know if i'm wrong and we should consider cases when the name in a layout conf and the actual JSON key (meaning `json_name` really matters).

nautics889:
@rkashinin

rkashinin:
Thanks for the review,
I think that you are right and in this situation json_name doesn’t matter, but there is no guarantee that json_name and layout_json always equal. That’s why without json_name now, it will be much more difficult to change the program later.

From my point of view it's a matter of design but the setitem function sets the element in the dictionary. So I don’t really understand, why do you prefer square brackets?

rkashinin:
[0001-Fix-for-issue-676-title-field-not-seem-to-be-working.patch](https://github.com/gnome-terminator/terminator/files/11186222/0001-Fix-for-issue-676-title-field-not-seem-to-be-working.patch)
[0002-Fix-for-issue-676-title-field-not-seem-to-be-working.patch](https://github.com/gnome-terminator/terminator/files/11186228/0002-Fix-for-issue-676-title-field-not-seem-to-be-working.patch)
Thanks for your explanation,
According to your review I’ve made some changes and I send you patch files with them.
Is it ok, or should I make another pull request?

nautics889:
@rkashinin thanks for the update!

You can just upload a new commit in a top of the current. Then it's going to be automatically update here, in the current PR.
So, it would be like this:
1. Open your project, ensure you're in the right branch (`issue-676` in your case): `git checkout issue-676`;
2. Make changes to code;
3. Make a new commit: `git add terminatorlib/configjson.py`, then `git commit` and a message like `minor fix`;
4. Upload then: `git push -u origin issue-676`;

(_Note:_ the argument "origin" in 4th step may be different depending on which settings are in your .git/config file; to clarify you can run `git remote -v` and chose the name which corresponds to URL of your fork on GitHub)
___
Alternatively you can create a new pull request, of course, it's up to you, choose the easier option :)
___
Hope this help!

rkashinin:
Thanks for your advice, 
I wasn’t sure about what to do, couse I’m a new one on git hub.
You helped a lot!

nautics889 (approved):

PR #738, opened 2023-04-04
fix: typo in french translation

Thibault-Brocheton:
dans = in
sans = without


mattrose:
my rudimentary knowledge of French from living within a km of Quebec agrees with you.

PR #747, opened 2023-04-22
Rewrite terminator_config.5 man page in AsciiDoc format

Vulcalien:
This took some time, but it's here!
In a previous PR (#733) I rewrote `terminator.1` in AsciiDoc. Now I've done the same for `terminator_config.5`.

The information in `terminator_config.5` was severely outdated and most of settings added through the years were missing.
I've also added a small script, `gen_manpages.sh`, that should be used to generate the manpages after changing the `.adoc` files.

### DIFF
--- a/doc/gen_manpages.sh
+++ b/doc/gen_manpages.sh
@@ -1,0 +1,4 @@
+#!/bin/sh
+
+asciidoctor -b manpage terminator.adoc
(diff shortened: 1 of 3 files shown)


mattrose:
Honestly didn't even review this.  I'm sure it's as good or better than anything I would have come up with.  Thanks again, @Vulcalien

Issue #744, opened 2023-04-19
Terminator prints raw (user input) root password while connected with sudo executing Terminator with `broadcast all` flag


serainox420:
#### About.
It prints raw out for root password if accidently ran some `sudo` if already not knowingly, connected to one terminator (and having any unsent char in another with auto `broadcast-all` that's gonna prevent `sudo` from running correctly in "broadcasted-to" terminal screen
It also logs it as raw into sys logs which is major concern...

#### How to replicate:

Open one terminator, insert any char (.,'[!@#$%^&* or literally any UTF-8 (and else) character)
Open second terminator (not knowing you left one in standby somewhere)
Do 'sudo' somethin in main terminator.
Check log in "background standby" terminal.

#### Suggested changes:
Ignore entire '\n' newline if `sudo` found in user input line,  no further than like 20chars from 0 (just to be sure no false positive det)
![terminatorpass](https://user-images.githubusercontent.com/95081005/232956956-6f8501d2-2bd2-4d62-9556-a885ca2a7c90.png)

nautics889:
@serainox420 could you please test the following PR https://github.com/gnome-terminator/terminator/pull/748?

PR #748, opened 2023-04-22
feat: Prevent broadcasting chars for passwrd #744

nautics889:
Issue #744 

When broadcasting option is enabled, check if the terminal actually expects a password before emitting the event ("key-press-event") to a specific terminal instance. If no, the event wouldn't be emitted to other terminal instances.

___

I couldn't find a better idea than check if there's password propmt at the cursor position (like when you run `$ sudo su -` then it prints "[sudo] password for username:")

### DIFF
--- a/terminatorlib/terminator.py
+++ b/terminatorlib/terminator.py
@@ -4,6 +4,8 @@
 
 import copy
 import os
+import re
+
 import gi
 gi.require_version('Vte', '2.91')
 from gi.repository import Gtk, Gdk, Vte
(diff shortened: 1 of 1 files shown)


serainox420 (approved):
Looks good to me 😎

serainox420:
Looks good.

nautics889:
@Vulcalien thank you for the feedback!

I totally agree with your points. And the same thoughts i had when i was investigating possible ways to achive this. Especially for false-positive cases when user wouldn't know why broadcasting doesn't work. Perharps, when the user leaves some characters in one of terminals in a group and types password in another one, reposibility for this should be lies with the user.

So i close the PR at this moment. Ping me if we return to this topic further.

PR #756, opened 2023-05-07
remove deprecated pytest-runner

gotmax23:
See https://github.com/pytest-dev/pytest-runner/#deprecation-notice. We can call pytest directly instead.


mattrose:
Not sure what this does, but I don't see any harm, and looks easily revertable if it does.

PR #767, opened 2023-06-04
add new tab after current tab option

x000zh:
Just a common option could be found in other terminal


mattrose:
Thank you very much for your help!

PR #777, opened 2023-06-18
Alter get_text_range call based on VTE version

### LINKED ISSUE #776: Logger plugin crashes with "TypeError: 'NoneType' object is not subscriptable"
Starting the "terminator" without the config file gives the same results.

Fill out as many of these fields as you can

**Describe the bug**
The logger plugin does not work with the new gnome-terminal (tested it on ubuntu 23.04 and on the rolling release of Arch), the plugin crashes with "TypeError: 'NoneType'"

**To Reproduce**
Start the logger, select the destination file, produce a terminal output (example: ps -Af), and stop the logger. The logger will fail to stop and no content is written to the output file.

**Expected behavior**
Termina output is written to output file

**Screenshots**
```bash
Error log from debug mode:
Window::on_window_state_changed: fullscreen=False, maximised=False
  File "/usr/lib/python3.11/site-packages/terminatorlib/plugins/logger.py", line 61, in save
    self.write_content(terminal, last_saved_row, last_saved_col, row, col)
  File "/usr/lib/python3.11/site-packages/terminatorlib/plugins/logger.py", line 48, in write_content
    fd.write(content[:-1])
             ~~~~~~~^^^^^
TypeError: 'NoneType' object is not subscriptable

```

 - Linux Ubuntu
 - Version 23.04
 - Display Technology:  Tested on both Wayland and X11 (ubuntu 23.04 wayland and arch was on X11)

mattrose:
In vte 0.72 there was a regression that caused Vte.Terminal.get_text_range() to fail to return the text requested However, in the same version, there was a new call introduced that does retrieve the text, so if I detect a vte minor version above 72, I call that instead.

Fixes #776

### DIFF
--- a/terminatorlib/plugins/logger.py
+++ b/terminatorlib/plugins/logger.py
@@ -6,7 +6,7 @@ terminals """
 
 import os
 import sys
-from gi.repository import Gtk
+from gi.repository import Gtk,Vte
 import terminatorlib.plugin as plugin
 from terminatorlib.translation import _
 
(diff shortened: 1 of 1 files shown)


mattrose:
Works on older vte too.  Merging

Issue #760, opened 2023-05-11
working directory feature is broken


tomatac:
Directory feature per terminal is ignored loading the config file and cannot be set using the UI.
Tested on Ubuntu 22.04.
This used to work in older versions.

The only reasonable workaround, until this gets fixed, is to use the custom commands plugin and setup there cd custom commands.

[config.txt](https://github.com/gnome-terminator/terminator/files/11447041/config.txt)

vssdeo:
Hi @tomatac 

This issue is on terminator 2.1.3 ? You mention that it used to work on earlier versions ? Can you please let me know what was the flow like and on which version of terminator ? 

So there is a plugin SaveLastSessionLayout, if enabled it will save your session as per the last window closed. On launch, you can Right Click->Layouts->SaveSessionLayout and restore that one. I have commented out the provision for auto loading as it wasn't sure how users would like to use it and if it was ok to override "default" so "SaveSessionLayout" was used. May be auto-restore should be a selectable option.

tomatac:
The issue is in version 2.1.3.
The feature is there in the UI. We can set a default folder per terminal. It is very useful especially when we test different microservices. We ran terminator configured with the correct folder for each service and is super useful to start multiple windows with everything ready to run.  
The issue is that it used to work and does not work anymore. 
Right click and go in "Preferences", then "Layouts" and select a terminal. There is an option to set a custom command and a working directory for each terminal. You can go through each terminal and add working directories and custom commands and it looks like they are retained but as soon as we press the "Save" button at the bottom left they are gone. I also tried to configure these settings in the config file but they get ignored. 
I used these features before and they used to work. 
SaveLastSessionLayout did not seam to preserve the last folder for each terminal window.
![image](https://github.com/gnome-terminator/terminator/assets/15964070/7ce0d06c-0894-47dd-9c6c-27a2bed3a3c7)

vssdeo:
Thanks for a detailed reply, Ok I think I get it, for 2.1.3, for "default" session

1. It seems that if you close the window without pressing "save",  the directory is saved and restored in next launch.
2. If you do press "save" then it does not save the directory.

For SaveLastSession. It works for me. One has to ensure that the window with tabs and working dir be closed last. Then on next launch right click->layout->SaveLastSession would open. May be it should be more intutive and by default should open last session when enabled.

vssdeo:
[...]

PR #804, opened 2023-08-06
[bug-760] working directory feature is broken 

vssdeo:
**Issue** Ref: https://github.com/gnome-terminator/terminator/issues/760

    - as per the bug the layout does not get updated when "Save" button in layout is pressed
    - it does get updated if the window is closed by pressing top x close icon.
    - on pressing Save, it seems that the prefseditor takes this from current_layout = terminator.describe_layout() and saves it in config
    - whereas the current changes are done per key stroke and config is updated
    - this patch copies the parameters like directory and command when Save is press and on_layoutrefreshbutton_clicked() is called
    - Hence working dir and command are copied when Save is pressed using uuid to match terminals.
    - If there is a command registered then the terminal runs the command and exits. so one is not able to see the results.

**Observation**

It was reported in bug that this used to work in previous versions, but I could not find that for me it only worked after the patch.

_command in layout issue:_

I have notices in the released version 2.1.3 and in the master branch I was working that if a command is present the terminator exits after being launched. I could trace it to terminal.py in following lines. I am not sure if thats the default behavior but it seemed confusing when program exited. 

elif self.config['exit_action'] in ('close', 'left'):
            self.cnxids.new(self.vte, 'child-exited',
                                            lambda x, y: self.emit('close-term'))

**Testing**

- Launch terminator
- Right-click and Preferences->Layout
- Try changing working directory and press Save below.
-  exit terminator and relaunch
- the previous state would be same unmodified

- after this patch that should work

### DIFF
--- a/terminatorlib/config.py
+++ b/terminatorlib/config.py
@@ -499,6 +499,34 @@ class Config(object):
         """Set a layout"""
         return(self.base.set_layout(layout, tree))
 
+    def copy_layout_item(self, src_layout, dst_layout, item):
+        items = {}
+        for child in src_layout:
+            section   = src_layout[child]
+            sec_type  = section.get('type', None)
+            if sec_type != 'Terminal':
+                continue
+
+            cp_item = section.get(item, None)
+            uuid    = str(section.get('uuid', None))
+            if cp_item:
+                items[uuid] = cp_item
+
+        dbg("items to be copied:%s" % items)
+        for child in dst_layout:
+            section   = dst_layout[child]
+            sec_type  = section.get('type', None)
+            if sec_type != 'Terminal':
+                continue
+
+            uuid       = str(section.get('uuid', None))
+            update_item = items.get(uuid, None)
+            if uuid and update_item:
+               dbg("update layout item:(%s) with value:(%s)"
+                                            % (item, update_item))
+               section[item] = update_item
+
+
 class ConfigBase(Borg):
     """Class to provide access to our user configuration"""
     loaded = None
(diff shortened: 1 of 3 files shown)


vssdeo:
Note: issue #760 update by the reporter notes that, it does not work for him though the update of directory and command is done, I am following up on that

vssdeo:
Let's wait for this one, cross-checking all conditions

mattrose:
@vssdeo I'm getting an error when I'm trying to merge this:
```

git checkout -b vssdeo-760-working-directory-feature-is-broken master
git pull https://github.com/vssdeo/terminator.git 760-working-directory-feature-is-broken
Switched to a new branch 'vssdeo-760-working-directory-feature-is-broken'
remote: Enumerating objects: 13, done.
remote: Counting objects: 100% (13/13), done.
remote: Compressing objects: 100% (5/5), done.
remote: Total 13 (delta 8), reused 10 (delta 8), pack-reused 0
Unpacking objects: 100% (13/13), 17.90 KiB | 1.12 MiB/s, done.
From https://github.com/vssdeo/terminator
 * branch              760-working-directory-feature-is-broken -> FETCH_HEAD
Auto-merging terminatorlib/config.py
Auto-merging terminatorlib/plugins/mousefree_url_handler.py
CONFLICT (add/add): Merge conflict in terminatorlib/plugins/mousefree_url_handler.py
Auto-merging terminatorlib/prefseditor.py
error: could not apply 0ae2aba1... 1) Mouse less/free opening / yanking feature
hint: Resolve all conflicts manually, mark them as resolved with
hint: "git add/rm <conflicted_files>", then run "git rebase --continue".
hint: You can instead skip this commit: run "git rebase --skip".
hint: To abort and get back to the state before "git rebase", run "git rebase --abort".
Could not apply 0ae2aba1... 1) Mouse less/free opening / yanking feature 2) plugin_util  file to assist this and other plugins with key-press Helper function 3) adding Key Binding Help Functions in above (2) to config and preferences->keybinding to have a consistent behavior
```
Can you fix this up so that it can be merged again?

PR #806, opened 2023-08-08
805 keyboard shortcut support for plugins code decoupling

vssdeo:


mattrose:
This looks good!  Thanks again, and sorry about the wait.

vssdeo:
@mattrose thanks for looking into this may be once back I should add documentation for key bind utility :) yes ‘documentation’ :) cheers !

Issue #808, opened 2023-08-09
Plugins don't receive keboard signals on newly opened windows


vssdeo:
Some plugins using KeyBindUtil https://github.com/gnome-terminator/terminator/issues/805 don't receive Key events in newly opened window / terminal. Reloading  of plugins helps and gives plugins an option to reconnect terminal events. This patch given on way or opens up a discussion on this and to test this.

**DEPENDENCY** 

Examples:

https://github.com/gnome-terminator/terminator/issues/706
https://github.com/gnome-terminator/terminator/pull/682

mattrose:
Assuming that the merged code fixes this.  Please reopen if this isn't the case

Issue #805, opened 2023-08-08
keyboard-shortcut-support-for-plugins-code-decoupling


vssdeo:
This bug/issue separates the KeyBind utility code where the plugins have the capability to register short cuts which can then be edited via preferences window. The issue 681 had this code and plugin both. Now this separates the utility code from plugin.

Details Ref:  functionalities of the plugin https://github.com/gnome-terminator/terminator/issues/681 along with screenshots

A separate issue is seen where the keyboard signals attached to window are not caught by new windows but work on multiple tabs of same window they work fine. I have tried to look into it and tried to address it by one approach. https://github.com/gnome-terminator/terminator/pull/682/commits/37c77161f2905c39780dc28c1eecd0f945d7f9ab

vssdeo:
Examples on how to use:
https://github.com/gnome-terminator/terminator/issues/706
https://github.com/gnome-terminator/terminator/pull/682

mattrose:
Assuming the merged code fixes this issue

Issue #706, opened 2023-01-28
Favorites/Bookmarks ???


Foul:
Hello, would it be possible to add Favorites/Bookmarks in this terminal ?

I'm using Konsole by default and i swould like this feature in Terminator plz.

Thx.

vssdeo:
Putting in some code to check things via Plugin.

**DEPENDENCIES:** KeyBind Util https://github.com/gnome-terminator/terminator/pull/806
**Known Issues:** 

-keyboard short cuts have issue on new windows, #681 puts in a solution to reload but that needs to be verified by others also. -GUI works.
-if two windows are opened, sometimes the second one while loading bookmark does not load the directory for the first time.

**Usage:** 

```
<Alt>b to add 
<Alt>r to remove
or right click and use via menu
```

- sub-menu via right click->bookmarks->urls->(act menu)

- bookmark plugin, basic features 
- Requires support of KeyBindUtil from Plugin which assists short cuts
- may be edit option  can be added via Preferences Plugin window 
- config has = so rhs can be used for tags
- TODO: config changes if can be brought to plugin code
```
-       support in Preferences window
-       limit on number of urls via Preferences ?
-       display shortcuts in menu
-       selection of urls from keyboard and tag support

```


![bookmark_plugin](https://github.com/gnome-terminator/terminator/assets/19357612/d6acd814-007b-4033-a355-31ebcc0beba3)

vssdeo:
Have changes in my branch Ref: https://github.com/vssdeo/terminator/tree/706-Favorites-Bookmarks-Plugin for testing and trying out.
- basic plugin
- tags feature
- UI support via prefs

![bookmark-plugin-2-tags](https://github.com/gnome-terminator/terminator/assets/19357612/3b8c5133-1933-4e86-8551-9f3173d4f017)
![bookmark-plugin-prefs-menu](https://github.com/gnome-terminator/terminator/assets/19357612/516d6023-0576-476b-bea5-4456d97f9e26)

vssdeo:
Hi. Have removed my old separate bookmark plugin and have added the functionality to custom command plugin. Would be great if anyone can test / try Ref: https://github.com/vssdeo/terminator/tree/706-Favorites-Bookmarks-Plugin 

Changes rest on my branch not putting up a pull request now. 

To test, just merge my changes or use branch. enable custom command plugin. Default keybindings are <Alt>b or direct adding <Shift><Aft>b to launch custom command pref window.

Current changes:     
[...]

PR #809, opened 2023-08-09
[bug 808] Plugins-dont-receive-keboard-signals-on-newly-opened-windows

vssdeo:
Ref: https://github.com/gnome-terminator/terminator/issues/808

Some plugins using KeyBindUtil https://github.com/gnome-terminator/terminator/issues/805 don't receive Key events in newly opened window / terminal. Reloading of plugins helps and gives plugins an option to reconnect terminal events. This patch given on way or opens up a discussion on this and to test this, may be a better solution exists.

DEPENDENCY

Examples:

https://github.com/gnome-terminator/terminator/issues/706
https://github.com/gnome-terminator/terminator/pull/682

### DIFF
--- a/terminatorlib/plugin.py
+++ b/terminatorlib/plugin.py
@@ -69,11 +69,13 @@ class PluginRegistry(borg.Borg):
         if not self.available_plugins:
             self.available_plugins = {}
 
-    def load_plugins(self):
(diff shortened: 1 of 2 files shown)


mattrose:
This looks good.  Thanks so much and sorry about the wait.

PR #816, opened 2023-08-26
Fix insert term name

mattrose:
Fix traceback when using "Insert Terminal Name"
    
    The "Insert terminal name" Plugin causes the following traceback
      File "/Users/mattrose/Code/terminator/terminatorlib/terminator.py", line 588, in do_insert_term_name
        term.feed(name)
      File "/Users/mattrose/Code/terminator/terminatorlib/terminal.py", line 1709, in feed
        self.vte.feed_child(text)
    
    This PR fixes it so the plugin inserts the terminal name, as intended.


mattrose:
Woah!  Too much!

Issue #802, opened 2023-07-24
Ability to undo or restore changes to the preferences


waldyrious:
**Is your feature request related to a problem? Please describe.**
I sometimes experiment with making changes in the Preferences window, to see how the various options look/work, and I may forget to memorize or back up the previous settings prior to the changes, which then leads me to be unsure how to restore the previous state.

**Describe the solution you'd like**
It would be awesome if there was an "Undo" button (and maybe a "Redo" button as well?) in the Preferences window, similar to the Back/Forward history buttons in browsers, text editors, etc. This would allow reverting the last change done in the current view of the Preferences window. There's no need to persist a full log of changes after the Preferences window is closed (or even as one changes between tabs of the preferences window), but both additions would be welcome as well! :D

**Describe alternatives you've considered**
Maybe a way to restore each preference (individually) to its default setting would be nice too. But I'm not sure how that could be implemented without cluttering the UI.

Another option would be an additional button next to "Close" in the bottom right of the Preferences button, labeled "Close without saving", or "Discard changes", or something like that. This would undo all the changes made since the Preferences window was opened.

PR #822, opened 2023-09-04
 [bug 802] - Ability to undo or restore changes to the preferences #802

vssdeo:
**Changes** 
- added basic config restore option from preferences window
 - changes to glade made in addition to addition of destory signal
 - minor cleaup in config to have a single function return confilg filename via get_config_filename

**Things to Cross Check**
- position of restore button in preference window
- name of button
- additional behavior like enabling button only on change or a simple standard way like current one is ok ?

### DIFF
--- a/terminatorlib/config.py
+++ b/terminatorlib/config.py
@@ -713,22 +713,55 @@ class ConfigBase(Borg):
         return filename
 
     def save_config_with_suffix(self, suffix):
-        filename = self.get_config_filename()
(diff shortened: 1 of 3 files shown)


mattrose:
I like it!  Thanks for your work and sorry about the wait

waldyrious:
Thanks for implementing this, @vssdeo! For future reference, below is a screenshot showing what this looks like. It consists of a "Discard Changes" button at the bottom of the preferences screen:

image

PR #823, opened 2023-09-04
706 favorites bookmarks plugin

vssdeo:
-basic bookmark support via plugin Ref:  Favorites/Bookmarks ??? #706


vssdeo:
I think lets wait a bit for this, I want to have a few more things in here, checking them out, else will delete the pull request

vssdeo:
removing pull request

PR #826, opened 2023-09-17
[bug 827] - toggle-visibility no longer opens a new window if terminator is running

gfickel:
A recent PR #797 added the toggle-visibility option to circumvent some wayland problems. There was a problem, though: it would also create a new Terminator window on each toggle. Tested on Ubuntu 23.04.

**The fix**

I simply changed the if/else position. On the original PR, `ipc.new_window_cmdline(optionslist)` would always by running `terminator --toggle-visibility`, and now a new window will never be created if --toggle-visibility is passed, as I believe would be the expected usage.


j-d-r:
Duplicate of PR: https://github.com/gnome-terminator/terminator/pull/824

j-d-r:
Your fix will not fix --new-tab

mattrose:
Closing as dupe of #824

Issue #830, opened 2023-09-29
Option for prompt for closing against accidental closure


H0R1Z0N:
**Is your feature request related to a problem? Please describe.**
I have often hard drive that i need to run long session of badblocks, often 40h+ session, i have a lot of shortcut on my mmorpg mouse and the mouse's in the bed where i sleep with my desktop, sometime, when handling my mouse to move it back when i wake or return in bed, i press the wrong shortcut and close terminator or press the x button on the taskbar when trying to switch screen session.

**Describe the solution you'd like**
I saw some 10yo post about an edit of a python 2 script of the terminator behavior for a closing prompt, which sadly python 2's not supported anymore for a while. Would it be possible to add the feature for a prompt if multiple windows tab or session are open and if at least a command's being run live in the terminal. I don't expect it to be too complicated to add.

**Describe alternatives you've considered**
Off course i could simply foreground/background each application to be sure they won't quit when i accidently press the wrong button, but that would not be conveniant.

**Additional context**
Since 2+ post on google give result about an edited python 2 script for terminator to add that behavior, i don't expect it to be too complicated to add.

I thank you for your great work, it's my favorite terminal emulator and i don't expect to try terminal hoping anytime soon, It's perfect when you need to do so much on a single screen.

H0R1Z0N:
I have no luck... I closed it again by accident and took a while to realise i did... Guess that will add another 20h of wait time on my long process i was doing...

Vulcalien:
We already have a similar feature, that asks confirmation before closing *multiple* terminals.

In the meanwhile, you could open another tab just to prevent from accidentally closing terminator, and make sure the setting is enabled. If it isn't enabled, go to the file `.config/terminator/config` and delete the line `suppress_multiple_term_dialog = True`.

But of course that's more than an hack than a solution. It shouldn't be too difficult to modify the existing feature to ask even if there is only one terminal.

mattrose:
Wrote a preferences gui for this as well

PR #834, opened 2023-10-10
Ask before closing, even if there is only one terminal

Vulcalien:
(Partially?) solves #830

This is apparently a very requested feature. This PR makes it possible for the "Close?" confirmation dialog to appear even if there is only one terminal.

## Settings: suppress_multiple_term_dialog ---> ask_before_closing
The `suppress_multiple_term_dialog` setting has been replaced with `ask_before_closing`, which can have values '_never_', '_multiple_terminals_' and '_always_'.
'_multiple_terminals_' is the default, so that Terminator's default behavior remains the same. The "Do not show this message next time" checkbox sets `ask_before_closing` to '_never_', regardless of its previous value.

## GUI changes
The confirm close dialog is the same as before, for multiple terminals. For one terminal, it is like this:
terminator-confirm-close

## Code changes
`Container.construct_confirm_close` has been significantly modified. The translation for "Close multiple terminals?" was broken and a few translatable texts were added.

### DIFF
--- a/doc/terminator_config.5
+++ b/doc/terminator_config.5
@@ -1,13 +1,13 @@
 '\" t
 .\"     Title: terminator_config
 .\"    Author: [see the "AUTHOR(S)" section]
-.\" Generator: Asciidoctor 2.0.18
-.\"      Date: 2023-04-22
+.\" Generator: Asciidoctor 2.0.16
+.\"      Date: 2023-10-10
 .\"    Manual: Manual for Terminator
 .\"    Source: Terminator
 .\"  Language: English
 .\"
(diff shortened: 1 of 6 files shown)


mattrose:
@Vulcalien you're amazing and so appreciated, even though I seem to show it by going months without touching Terminator.  I'll take a look.

mattrose:
Also love the fact that you fixed docs at the same time.

Vulcalien:
No worries about the inactive time, we're all volunteers after all.

mattrose:
Merging this.  

I worked out the Glade for adding a preferences entry for this as well. Now I just have to wire it up.

PR #841, opened 2023-10-16
706 favorites bookmarks plugin

vssdeo:
- modified custom command plugin to have bookmarks
 - made name parsing of the custom command optional, so whole command can be added instead of the command being auto split into sub-menus. Default behavior is kept and a checkbox option is provided.
 - added keybindings for directly adding or launch the custom command menu
 - the dialog window taking command for bookmark via shortcut keybinding will have focus on OK Button for faster interaction
 
After enabling plugin KeyBindings available for fast adding custom command `<Alt>b` to launch custom command pref menu `<Shift><Alt>b`


mattrose:
Looks good, thank you!

Issue #773, opened 2023-06-12
Feature: is there any way to  Hide or Sort context menu items?


yurenchen000:
the built-in menu, and plugins added menu item.

there can be dozens of menu items.
Is there a way to 
- **Hide** menu item
  - OR categorized into the **secondary menu** //like the `Grouping` menu in 2.1.3
- **Sort** menu item by custom order
  -  OR split line into different item **group** // like the current `Copy` `Paste` item followed by a horizon line

------------

Is it possible implement it in a plugin,
OR have to change the terminator source code?

Vulcalien:
This is something that should be implemented. I also found myself overwhelmed by all the options and it's only going to get worse over time, as we add new ones.

Currently, only a few can be disabled (without xhanging the code) by disabling the plugin that generates them.

yurenchen000:
>  I also found myself overwhelmed by all the options and it's only going to get worse over time, as we add new ones.

Can't agree more.

Purify and classify configuration items to achieve a simple and smooth setting UI?
// Put less commonly used settings in configuration files or advanced settings ?

<br>

Tabby[^1] has a smooth settings experience.

[^1]: https://github.com/Eugeny/tabby

yurenchen000:
> Currently, only a few can be disabled (without changing the code) by disabling the plugin that generates them.

And plugins no there own setting page,
They have to put setting entry into contenxt menu.

mattrose:
Two things about this.  We should put all plugin context menu items in their own Plugins sub-menu.

We should look to see if there's a way to integrate plugin configurations into the plugin tab of the preferences section.  There's room for it there.

@yurenchen000 can you say more about what you like for tabby in particular?  Is the context menu configurable somehow?

yurenchen000:
apologize for the late reply, it's been a busy week.
(I'm not a UX engineer,  just a heavy everyday user.

> We should put all plugin context menu items in their own Plugins sub-menu.

I thought that plugins declare/create menu, but end user use it. 
so maybe give the choice to user for customize context-menu by show/hide sort/group items in top-level context-menu.

Something like custom extension toolbar in chrome(and firefox)
// I thought the context menu in terminator was as important as the browser toolbar, but it's easy to get messy
![chrome_toobar_customize-](https://github.com/gnome-terminator/terminator/assets/8458213/a2de51b4-5137-489d-aed6-9391742d2ee3)

yurenchen000:
[...]

PR #842, opened 2023-10-16
[Plugin ContextMenuPlugin] 773-Feature-is-there-any-way-to-Hide-or-Sort-context-menu-i…

vssdeo:
**Context Menu Plugin**

**Note:** Plugin affects core / basic features so up for testing, suggestions and tweaks.

-It's ready for review and comments but since it may require tweaks and other changes so putting up the changes.

Ref: For Testing and Details https://github.com/gnome-terminator/terminator/issues/773

- Adds a context menu (right click)

- Gives an UI under Preferences->Plugins with capability to organize items
 
- Supported by PluginEventRegistry which helps to add plugin actions to menu
- apart from other plugins using this, for eg. this context_menu_plugin can
- register its menu functions while itself creating a menu and so can other plugins
 
- Supported by KeyBindUtil for action key / desc matching
 
- Cleans and identifies common dependencies which can be further worked on
 
- Gradual removal of menuitems to be done with checking logic and removal of if
- based conditions. All cases are being compiled in class Filter below which have to be removed
 
- Menu Names are taken from keybinding name

### DIFF
--- a/terminatorlib/config.py
+++ b/terminatorlib/config.py
@@ -123,6 +123,10 @@ DEFAULTS = {
             'new_tab_after_current_tab': False,
         },
         'keybindings': {
+            'zoom'             : '',
+            'unzoom'           : '',
+            'maximise'         : '',
+            'open_debug_tab'   : '',
             'zoom_in'          : '<Control>plus',
             'zoom_out'         : '<Control>minus',
             'zoom_normal'      : '<Control>0',
--- a/terminatorlib/plugin.py
+++ b/terminatorlib/plugin.py
@@ -30,6 +30,8 @@ from .config import Config
 from .util import dbg, err, get_config_dir
 from .terminator import Terminator
 
+from .version import APP_NAME
+
 class Plugin(object):
     """Definition of our base plugin class"""
     capabilities = None
@@ -117,6 +119,11 @@ class PluginRegistry(borg.Borg):
 failed: %s' % (plugin, ex))
 
         self.done = True
+
+    def get_plugin_instance(self, plugin):
+        instance = self.instances.get(plugin, None)
+        dbg('get plugin: %s instance: %s' % (plugin, instance))
+        return instance
 
     def get_plugins_by_capability(self, capability):
         """Return a list of plugins with a particular capability"""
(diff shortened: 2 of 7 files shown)


mattrose:
This is a pretty massive change to the context menu, something which was relatively simple up to now.  I've been very careful not to add too many items to the context menu, for fear of it getting too big.  I still don't think it's too big, when compared to other context menus, so I'm not sure what to do with this.  I don't know that it's necessary today, but it might become necessary in the future.  Also, GTK4 **drastically** changes signal handling, which seems to be a big part of this, and a lot of it might have to be thrown out.  

I'm definitely not closing this right now, but I will continue to play with it.

I'm not sure if it's easy to break up into smaller, more manage-able chunks, but you might look at that as well.

vssdeo:
Hi @mattrose, first thanks for taking your time and looking into this. All wise suggestions & helpful as always.Yes, playing around with this enabled would be a good idea.

Even I was under the idea that how to approach, so the thought process was to be consistent in approach and open things for discussion. For a brief idea I have listed down changes below, that helps to understand the code changes.

- I thought that if the change is submitted as a block people would be able to test with a single pull request & discard from their branch.
- I have kept the plugin functionality separate.
- The only thing that overlaps is the interface where keybindings are merged, which helps to simplify the prev code in prefseditor.py, this was also put in here so the change is visible and shows its integration and the change in terminal.py where we detect non core actions.

The point about not expanding the context menu is also wise. I will think more on this or may be in the plugin addition of new elements can be restricted to those defined in context menu plugin and those who are hacker itchy enough can enable in if needed :-P

I will further check if something more can be decoupled.


**Code-Change-Summary**

config.py  -   only adds actions to config to make things consistent

terminal_popup_menu.py - only old code commented out, nothing added

terminal.py    - minimal changes to detect the keypress / short-cut 

plugin.py       - minimal changes to prev code for consistency
                      - new section for the context menu event 
                      - new section for gui interface

prefseditor.py - added actions as per config
                       - changes made to clean up actions as per plugin.py
                       - new section for gui interface

plugins/PluginContextMenu.glad - new separate glade file
plugins/context_menu.py             - new separate plugin file

Issue #843, opened 2023-10-17
[Plugin SaveLastSessionLayout] not saving layout when user logs out or shutdown or restart


vssdeo:
Plugin SaveLastSessionLayout not saving layout when user logs out or shutdown or restart

- Have terminator its windows and tabs open.
- Log out or Shutdown or kill process
- On return, (Right Click)->Layout->SaveLastSessionLayout

- It won't restore the layout

**Expected behavior**

-It should restore the layout

PR #844, opened 2023-10-17
[bug 843] 843-Plugin-SaveLastSessionLayout-not-saving-layout-when-use…

vssdeo:
Ref: https://github.com/gnome-terminator/terminator/issues/843
- fixed Plugin SaveLastSessionLayout not saving layout when user logs out or shutdown or restart
- signals added

### DIFF
--- a/terminatorlib/plugins/save_last_session_layout.py
+++ b/terminatorlib/plugins/save_last_session_layout.py
@@ -1,4 +1,5 @@
 import os
+import signal
 import sys 
 
 # Fix imports when testing this file directly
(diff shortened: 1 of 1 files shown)


mattrose:
@vssdeo I tried this and this didn't fix the issue, as far as I could tell.  Can you double-check this with an empty config file?  I'll give it another stab soon, but I wanted to let you know

vssdeo:
@mattrose with an empty config file the session save plugin won't be enabled so it won't work, can you try with the plugin enabled. For that the config file needs to be in place. Let me know and I will check again if our results vary.

PR #848, opened 2023-10-20
New plugin: additional key shortcut [Ctrl+Shift+L] to clear all terminals at once

jachoo:
This simple plugin adds a new keyboard shortcut [Ctrl+Shift+L] which clears all visible terminals by sending [Ctrl+L] to each terminal.


mattrose:
Confined to plugin code:

Good Things:  Isolated
Bad things:  Unable to check to see whether Ctrl-Shift-L is already taken by another keybind.   Do you know what happens in this case?

jachoo:
If a key is already bound, it will be overridden, which could potentially lead to issues when unbinding. This is a general problem of binding keys in current implementation of `plugin.KeyBindUtil`

PR #856, opened 2023-10-26
[bug 852] - 852-terminator_py_get_focussed_terminal_always_returns_none

vssdeo:
- added get_vte().has_focus() for a valid focussed terminal to return


mattrose:
Thanks!

PR #870, opened 2024-01-11
Add support for a second keybind per action

### LINKED ISSUE #371: Multiple Keybinds for one Action
So, I split my terminal via the shortcut super key + right arrow key for a vertical split and super + arrow key down for a horizontal one. The problem is that splitting the terminal doesn't work with my left/upper key. 

My desired solution would be that for each action you could declare multiple keybinds.

I have considered setting up an external macro for super + left to trigger super + right, but that isn't a good idea, because it could interfere with other things and would have me run external software in the background.

rcalixte:
Modifying code:
 - terminatorlib/config.py
 - terminatorlib/keybindings.py
 - terminatorlib/plugin.py
 - terminatorlib/preferences.glade
 - terminatorlib/prefseditor.py
 - terminatorlib/terminal_popup_menu.py

Also updating translation files

Closes #371

**NOTES**
* This can be expanded to three bindings per action if needed but I thought two would be sufficient.
* Whitespace was cleaned up in the modified files in addition to addenda below.

**ADDENDA**
* Added missing parameters to base definitions
* Removed unnecessary parentheses from `return` and `del` statements
* Formatted regular strings that could be f-strings
* Various other code quality cleanups

cc @mattrose This will probably be easier to review with whitespace changes hidden. For the translation files, I just ran both scripts to pick up the new strings. I've tested this locally off of the master branch as well. For configs in the "old"/current format, there are config parsing warnings at startup (only visible if you launch from a terminal), but nothing visible otherwise to the user and nothing that impacts usability. Once any setting is modified, the new format for bindings will be written to the config and the warnings won't be seen again. So users who don't make any settings changes won't see anything change in their experience. Feel free to let me know if any additional changes are needed in order to get this merged and released.

(Side note: I'd only recently learned about Terminator and it's literally been the biggest boon for my productivity of late, immediately becoming my default on every machine. Thanks to you and the team for resurrecting and maintaining it!)


mattrose:
Wow, there's a lot here, I'll take a look as soon as I can.

mattrose:
So, there's so much here that it's impossible for me to do a review.  70,000 changes is way too big for a code review. The idea is good, but it looks like you went through it with a code formatter or linter that made a bunch of irrelevant changes.  That's fine as it is, but it makes it impossible for me to easily review the changes that are necessary to the feature, and the output of the linter.  Also, the doc changes should probably be done in a separate PR as well, so that we can more easily tease those out.

If you could make another PR with just the changes that are necessary to implement the feature, I could definitely review that and merge it in, with some work, but as it is, I can't really review or merge this PR.

rcalixte:
> If you could make another PR with just the changes that are necessary to implement the feature, I could definitely review that and merge it in, with some work, but as it is, I can't really review or merge this PR.

I mentioned that it would be easier to review with GitHub's whitespace option for that reason. The line change count is also largely due to the modified translation files due to the changed strings. I can pull those and the PR should be easier to review but I do think proper translation support should be implemented at some point in time.

rcalixte:
> **Question:**
> Should the workflows be updated for python 3.10 and 3.11? They are currently 3.8 and 3.9. I tested 3.10 and 3.11 on my local fork without issue so it should be safe to make that change. (Adding 3.12 would require working around the removal of packages like `setuptools` and I reason that isn't on the roadmap yet which is why I excluded it.)

Should I open the PRs for this too? We could add 3.10 and 3.11 while dropping 3.8. (I don't know of any distros that ship 3.8 **and** an updated terminator package.)

terminatorlib/keybindings.py
```diff
@@ -16,19 +16,23 @@
...
 
-Validator and functions for dealing with Terminator's customisable 
+Validator and functions for dealing with Terminator's customisable
 keyboard shortcuts.
 
 """
 
 import re
-from gi.repository import Gtk, Gdk
+from gi.repository import Gdk
```
mattrose, line 25: Why remove GTK here?  Is it not necessary?

terminatorlib/keybindings.py
```diff
@@ -91,7 +95,7 @@ def reload(self):
                     self._lookup[mask][keyval] = action
                     self._masks |= mask
 
-    def _parsebinding(self, binding):
```
mattrose, line 94: Why change this method name?

terminatorlib/keybindings.py
```diff
@@ -103,27 +107,25 @@ def _parsebinding(self, binding):
             raise KeymapError('No key found')
         keyval = Gdk.keyval_from_name(key)
         if keyval == 0:
-            raise KeymapError("Key '%s' is unrecognised" % key)
+            raise KeymapError(f"Key '{key}' is unrecognised")
```
mattrose, line 110: Why are you changing this string formatting?

terminatorlib/keybindings.py
```diff
@@ -103,27 +107,25 @@ def _parsebinding(self, binding):
             raise KeymapError('No key found')
         keyval = Gdk.keyval_from_name(key)
         if keyval == 0:
-            raise KeymapError("Key '%s' is unrecognised" % key)
```
mattrose, line 106: Why are you changing this string formatting?

terminatorlib/keybindings.py
```diff
@@ -103,27 +107,25 @@ def _parsebinding(self, binding):
...
+            raise KeymapError(f"Key '{key}' is unrecognised")
         return (keyval, mask)
 
     def _lookup_modifier(self, modifier):
         """Map modifier names to gtk values"""
         try:
             return self.modifiers[modifier.lower()]
-        except KeyError:
-            raise KeymapError("Unhandled modifier '<%s>'" % modifier)
+        except KeyError as exc:
```
mattrose, line 117: Why are you changing this string formatting?

terminatorlib/keybindings.py
rcalixte: It's an unused import here.

rcalixte: Should we remove the comment too?

mattrose: I'd rather keep code cleanup changes separate from feature or bugfix changes.  It means when I go back theough the history I can tell exactly what changes did what.  This is the point I was trying to get across with my earlier comment.  There's too much going on in this commit.  The stuff may be good to include in another PR, but in a PR that is just supposed to implement a feature, the code should just reflect what is necessary to implement that feature.   I definitely welcome bugfixes and code cleanup PRs, just not mixed in with feature implementation PRs

mattrose (commented):
I'd rather keep code cleanup changes separate from feature or bugfix changes. It means when I go back theough the history I can tell exactly what changes did what. This is the point I was trying to get across with my earlier comment. There's too much going on in this commit. The stuff may be good to include in another PR, but in a PR that is just supposed to implement a feature, the code should just reflect what is necessary to implement that feature. I definitely welcome bugfixes and code cleanup PRs, just not mixed in with feature implementation PRs

terminatorlib/keybindings.py
rcalixte: The underscore kept it as an internal method. This affected the tests from being able to access the method.

terminatorlib/keybindings.py
rcalixte: F-strings have been preferred since Python 3.6 went GA.

terminatorlib/keybindings.py
rcalixte: F-strings have been preferred since Python 3.6 went GA.

Issue #879, opened 2024-02-03
[Feature Request] Save also current python virtual environment in config file with SaveUserSessionLayout plugin


markdoerr:
**Is your feature request related to a problem? Please describe.**
Currently in the "SaveUserSessionLayout" only the current directory is stored in the config file.
I would be nice to also add also the current active virtual environments - and restore them later on. 

**Describe the solution you'd like**
If a terminal is using a virtual environment, also save it to the session config file.
Use this information to re-activate the virtual environment on session restore.

PR #880, opened 2024-02-03
WIP: Feature/save user session venv (#879)

markdoerr:
Additionally to the current working directory, it would be nice to also store the python virtual environment that is currently used in the config file and re-activate it on Session restore.

This solves #879

### DIFF
--- a/terminatorlib/plugins/save_user_session_layout.py
+++ b/terminatorlib/plugins/save_user_session_layout.py
@@ -1,0 +1,69 @@
+import os
+import sys 
+
+# Fix imports when testing this file directly
+if __name__ == '__main__':
+  sys.path.append( os.path.join(os.path.dirname(__file__), "../.."))
+
+from gi.repository import Gtk,Vte
+
+from terminatorlib.config import Config
+import terminatorlib.plugin as plugin
+from terminatorlib.translation import _
+from terminatorlib.util import get_config_dir, err, dbg, gerr
+from terminatorlib.terminator import Terminator
+from terminatorlib import util
+
+
+# AVAILABLE must contain a list of all the classes that you want exposed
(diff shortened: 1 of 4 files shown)


markdoerr:
Hi terminator developers,

I just need to figure out, how to get the virtual environment of a current terminal.

```
prefix = getattr(sys, "base_prefix", None) or getattr(sys, "real_prefix", None) or sys.prefix
     if prefix != sys.prefix: # session is in a virtual environment
         return sys.prefix
``` 
cannot work in this case, as we would need to run a python program within the current shell.

Do you know, if this information is kept somewhere in the shell / process ? (it should, since command prompts also show the current venv)

mattrose:
Can you fix the conflicts from your version change?  Then I think I can merge this in.  This is getting really close to adding functionality that is not strictly part of the terminal emulator process, but I think I'll allow it.

To really add venv functionality, you could look at adding venv to the layout editor as well.  Add it as an extra parameter to the layouts.

I'm worried that this is getting into "Python IDE" territory, but I'm allowing it as this is what a lot of people are actually using it for.

markdoerr:
Hi @mattrose, I can fully understand your concerns 
( when I started this branch, I also thought about it ;), but as it is a python based application and many people use python, it is worth having this feature included). 
So thanks for supporting this PR ! :)

There is one "game-stopper" preventing me from finallising the pull request (next to my time limitations):
I have not found an elegant, pythonic way to get hold of the name / path of the current virtual environment of a terminator window:
to my understanding of terminator, this is information is hold only by the shell running in the window. 
Is there a generic interface, I could use to query the shell for the python environment ? How can i run this code snippet shown above ? 
Do you have any idea ? Thanks, mark

PR #882, opened 2024-02-05
[bug 760] working directory feature is broken

vssdeo:
- as per the bug the layout does not get updated when "Save" button in layout is pressed
- it does get updated if the window is closed by pressing top x close icon.
- on pressing Save, it seems that the prefseditor takes this from current_layout = terminator.describe_layout() and saves it in config
- whereas the current changes are done per key stroke and config is updated
- this patch copies the parameters like directory and command when Save is press and on_layoutrefreshbutton_clicked() is called
- Hence working dir and command are copied when Save is pressed using uuid to match terminals.
- If there is a command registered then the terminal runs the command and exits. so one is not able to see the results.

### DIFF
--- a/terminatorlib/config.py
+++ b/terminatorlib/config.py
@@ -499,6 +499,34 @@ class Config(object):
         """Set a layout"""
         return(self.base.set_layout(layout, tree))
 
+    def copy_layout_item(self, src_layout, dst_layout, item):
+        items = {}
+        for child in src_layout:
+            section   = src_layout[child]
+            sec_type  = section.get('type', None)
+            if sec_type != 'Terminal':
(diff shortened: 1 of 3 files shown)


mattrose:
You did it you mad genius! :)

This has been a problem for so long it's in the docs that were written before I took over.  I'll take another look at it before merging it, but all I have to say is:  Wow, good job!

vssdeo:
> You did it you mad genius! :)
> 
> This has been a problem for so long it's in the docs that were written before I took over. I'll take another look at it before merging it, but all I have to say is: Wow, good job!

I shall print and frame this :)  @mattrose

PR #883, opened 2024-02-06
window.py: Fix window group toggle keybind

rcalixte:
The current conditional always evaluates to False, resulting in the ungroup_win() function never being called

### DIFF
--- a/terminatorlib/window.py
+++ b/terminatorlib/window.py
@@ -768,7 +768,7 @@ class Window(Container, Gtk.Window):
         # change
         child.set_current_page(child.get_current_page())
 
-    def set_groups(self, new_group,  term_list):
+    def set_groups(self, new_group, term_list):
         """Set terminals in term_list to new_group"""
         for terminal in term_list:
             terminal.set_group(None, new_group)
@@ -802,7 +802,7 @@ class Window(Container, Gtk.Window):
 
     def group_win_toggle(self, widget):
         """Toggle grouping to all windows in the current window"""
-        if widget.group == 'Window':
+        if widget.group:
             self.ungroup_win(widget)
         else:
             self.group_win(widget)


mattrose:
Looking at the commit, that will ungroup the window if the window is any string, but it specifically has to be 'Window'.  Unless there's something I'm missing.

rcalixte:
> Looking at the commit, that will ungroup the window if the window is any string, but it specifically has to be 'Window'. Unless there's something I'm missing.

That's what I'd done originally but if you create a new group using the menu, it defaults to a letter from the Greek alphabet. It also would miss the condition of custom window group titles. My understanding is that the intended function is to unset a group if one is currently set when the toggle action is triggered.

mattrose:
Hmm, I will poke, gimme a few minutes

mattrose:
The "Group All in Window" and "Ungroup All in Window" works as I would expect it.  When Selected, it creates a group named "Window Group x" where x is a number starting from 1.

When you select "New Group" It creates a new group named after a greek letter, and then you need to add all the rest of the terminals in that window to the group by hand.  Unfortunately, Terminator doesn't keep track of this as a Window group, and so the "Ungroup all in Window" doesn't behave as you expect.

If you want to use "Ungroup All in Window, you first need to create a Window Group.

rcalixte:
Adding a screenshot so we can start from the same page. 😃

rcalixte:
Are you saying that the two left terminals here are not considered a window group? If I hit the keybinding to toggle the window grouping, I would expect it to ungroup these windows, not create "Window group 2" for all the terminals. The same with the grouped terminals on the right.

rcalixte:
> Unfortunately, Terminator doesn't keep track of this as a Window group, and so the "Ungroup all in Window" doesn't behave as you expect.

I just tested this using the screenshot above as a base and it does indeed work as I expect. It ungrouped all of the terminals in the window, both "Lambda" and "Window group 1" in this instance.

mattrose:
Are you hitting the group_win_toggle keybind, or are you going up to the "Ungroup All in Window" menu item in the titlebar menu.

They ... do different things apparently

rcalixte:
> Are you hitting the group_win_toggle keybind, or are you going up to the "Ungroup All in Window" menu item in the titlebar menu.

I hit the menu item just now to make sure that it worked as expected. It did. The group_win_toggle keybind as it currently is would instead have created a new group named "Window group 2" if I recall correctly. (I can set it up the same way and then verify that.)

> They ... do different things apparently

That's what this PR is trying to fix. 😉

mattrose:
Thanks for working through this with me on gitter.im, @rcalixte

PR #903, opened 2024-04-19
Add reverse search keyboard binding

f3rn4nd0-c354r:
Add keyboard binding for reverse search (<kbd>Shift</kbd> + <kbd>Return</kbd>)


mattrose:
Love it.  Thanks!

PR #924, opened 2024-06-19
Add plugin to clone ssh and container sessions when terminal is split

asifamin13:
# Remote Plugin

A Terminator plugin which adds features for ssh and docker/podman to the context menu. 

Self contained and only relies on `psutil`. 

From the [terminator docs online](https://gnome-terminator.readthedocs.io/en/latest/plugins.html)
> Probably the most useful thing would be something to help customise Terminator for heavy ssh users

Alt Text

## Clone Horizontally/Vertically

This will clone your current SSH/container session into a newly spawned terminal

Heavily inspired by [terminator_plugins](https://github.com/ilgarm/terminator_plugins) which is no longer maintained. My version doesn't need to subclass `Terminal` and only needs `vte.feed_child`

## Profile Host Matching

When you clone a remote session, you can apply a terminator profile based on host or container name

Inspired by [TerminatorHostWatch](https://github.com/GratefulTony/TerminatorHostWatch) which does this via regex matching a PS1. Instead of using a regex my version uses the process info provided via psutil

## Configuration

Plugin section in `~/.config/terminator/config` :
```
[plugins]
  [[Remote]]
    # Automatically clone when you split a terminal with a remote session
    auto_clone = False

    # When a terminal with a remote session is cloned, attempt to parse the
    # current working directory via the PS1 and 'cd' into it
    infer_cwd = True

    # Optional default profile for all SSH sessions
    ssh_default_profile = common_ssh_profile

    # Optional default profile for all container sessions
    container_default_profile = common_docker_profile

    # You can override above defaults by specifing a host with a profile key
    # ex:
    [[[foo]]]
      profile = foo_profile
```

### DIFF
--- a/terminatorlib/plugins/remote.py
+++ b/terminatorlib/plugins/remote.py
@@ -171,7 +171,6 @@ class ContainerSession(RemoteSession):
 
     def GetHost(self, proc):
         """ try to find container name from cmdline """
-        # TODO: figure this out
         cmd = self._get_command(proc)
         if not cmd:
             return None
--- a/terminatorlib/plugins/remote.py
+++ b/terminatorlib/plugins/remote.py
@@ -66,7 +66,10 @@ import threading
 
 from typing import Optional, List
 
+import gi
 from gi.repository import Gtk, GLib
+gi.require_version('Vte', '2.91')
+from gi.repository import Vte
 
 from terminatorlib.plugin import MenuItem
 from terminatorlib.config import Config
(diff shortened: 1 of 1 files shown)


mattrose:
At first glance this looks like something that would be really great.  I'm going to pull it down and play with it for a bit just to see how to use it, but I can't see any reason not to merge it.  I'm very impressed that you were able to do this all within the plugin system.

asifamin13:
@mattrose thanks! let me know if you have any suggestions or improvements

mattrose:
I'm getting this error when I try to clone with split.
```
  File "/Users/mattrose/Code/terminator/terminatorlib/plugins/remote.py", line 680, in _split_axis
    self.remote_cwd = self._get_cwd_from_lines(terminal)
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/mattrose/Code/terminator/terminatorlib/plugins/remote.py", line 499, in _get_cwd_from_lines
    matches = list(self.cwd_regex.finditer(lines))
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: expected string or bytes-like object, got 'NoneType'
```
and the prompt is just
`mattrose@fiesole:~$ `

Maybe it's having trouble interpreting the ~ ?

mattrose:
Nope, it's another VTE Deprecation.  vte.get_text_range() starts returning null in 0.76, and you have to use the new as of 0.72 vte.get_text_range_format(), call.
Which means that we have to detect which version of vte the user is running and make the right method call.
I have some idea on how to do this but I haven't coded it yet.  Would you like me to continue, or do you want to code this up?

mattrose:
Either way, I'm keeping this, this is a great feature, coded very well, thank you.

asifamin13:
Ahh good catch! You can go ahead and patch it if you feel up to it. When I get back home tonight I can fix it also

mattrose:
easier if you do it, then you can just push the new changes to your branch and they should get automatically added to the PR.

asifamin13:
@mattrose just pushed the patch. It will check if vte >= 0.72 and call `get_text_range_format` appropriately. I tested this on Ubuntu 24 and 22 and works as expected either way.

Let me know what you think!

mattrose:
That patch looks great, pulling it in now.

PR #926, opened 2024-06-25
fix translation in popup menu

lexdene:
The texts in popup menu are not translated correctly.

The gettext function should be placed together with the original text,
or it will not be detected by `./genpot.sh`.

I think this bug is introduced by #663.


mattrose:
Looks good, thanks for the fix!

PR #965, opened 2024-11-28
[bug fix]: Fix screen-splitting issues in terminatorlib/terminal.py

0xh3xa:
**What:** 

A problem occurs when attempting to split the screen either horizontally or vertically in Terminator.

**_Traceback_:**

```
  File "/usr/lib/python3.13/site-packages/terminatorlib/terminal_popup_menu.py", line 198, in <lambda>
    self.terminal.get_cwd()))
    ~~~~~~~~~~~~~~~~~~~~~^^
  File "/usr/lib/python3.13/site-packages/terminatorlib/terminal.py", line 261, in get_cwd
    return(GLib.filename_from_uri(vte_cwd)[0])
           ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^
gi.repository.GLib.GError: g_convert_error: The hostname of the URI  “file://test-account/home/main” is invalid (4)
```

**Description of the fix:**

- This fix resolves an issue in `terminatorlib/terminal.py` where the improper handling of the `GLib.filename_from_uri` function caused the split-screen feature to fail due to an incorrect return value. 

- I've updated the `get_cwd(self)` function to use Python's built-in method for retrieving the current working directory instead of relying on `GLib.filename_from_uri`.

**Screenshots:**

image

A problem occurs when attempting to split the screen either horizontally or vertically in Terminator.

### DIFF
--- a/terminatorlib/terminal.py
+++ b/terminatorlib/terminal.py
@@ -254,10 +254,9 @@ class Terminal(Gtk.VBox):
 
     def get_cwd(self):
         """Return our cwd"""
-        vte_cwd = self.vte.get_current_directory_uri()
-        if vte_cwd:
-            # OSC7 pwd gives an answer
-            return(GLib.filename_from_uri(vte_cwd)[0])
+        cwd = os.path.expanduser('~')
+        if cwd:
+            return(cwd)
         else:
             # Fall back to old gtk2 method
             dbg('calling get_pid_cwd')


mattrose:
This doesn't duplicate the functionality.  It just sets cwd as the home directory.  Can you be more clear on the errors that you're seeing

mattrose:
@egmontkob has come up with a better fix

Issue #939, opened 2024-08-09
-x/--execute still broken


ManuelSchneid3r:
Same as https://github.com/gnome-terminator/terminator/issues/702

Especially as @vmath3us stated since there will be the XTE spec (hopefully) soon.

```sh
terminator -x echo -- hi  # works
terminator -x /bin/zsh -i -c "ls  ; exec sh"  # works
terminator -u -x /bin/zsh -i -c "ls  ; exec sh"  # works, but -u should not be required
```

Issue #923, opened 2024-06-19
terminator -x gets confused by --


meator:
**Describe the bug**
terminator tries to handle `--` after `-x` is specified, which is incorrect.

**To Reproduce**
```
terminator -x echo -- a
```

**Expected behavior**
Program working.

Instead, this happens:

```
usage: terminator [-h] [-R] [-v] [-m] [-M] [-f] [-b] [-H] [-T FORCEDTITLE] [--geometry GEOMETRY] [-e COMMAND] [-g CONFIG] [-j CONFIGJSON] [-x ...]
                  [--working-directory DIR] [-i FORCEDICON] [-r ROLE] [-l LAYOUT] [-s] [-p PROFILE] [-u] [-d] [--debug-classes DEBUG_CLASSES]
                  [--debug-methods DEBUG_METHODS] [--new-tab] [--toggle-visibility] [--unhide] [--list-profiles] [--list-layouts]
terminator: error: unrecognized arguments: -- a
```

Everything after `-x` should be `exec`ed verbatim without any additional argument parsing/handling.

The manpage says:

> Runs **the rest of the command line** instead of your default shell or profile specified command.

Special handling of `--` breaks this promise.

**Desktop:**
 - OS or Linux Distribution: Void Linux
 - Version: rolling release
 - Display Technology: X11
 - terminator version: reproducible on `2.1.3`, `2.1.4` and on `master`

PR #967, opened 2024-12-08
- [bug 678] fix argument handling of the --execute flag

vssdeo:
- [bug 939] execute still broken
- [bug 923] terminator -x gets confused by

- fixed double dash -- by handling extra arguments, need to cross check against all use cases
- currently its done in terminal.py need to check if other dependencies but first I want to get
- the behavior right

As per #939 #923 the use cases need to be further tested since I am not aware of previous behavior .

terminator -x echo -- hi  # fails
terminator -x /bin/zsh -i -c "ls; exec $SHELL"  # fails
terminator -u -x /bin/zsh -i -c "ls; exec $SHELL"  # works, but -u should not be required

### DIFF
--- a/terminatorlib/optionparse.py
+++ b/terminatorlib/optionparse.py
@@ -114,7 +114,7 @@ icon for the window (by file or name)'))
                 help=argparse.SUPPRESS)
 
     global options
-    options = parser.parse_args()
+    options, extra = parser.parse_known_args()
 
     if options.version:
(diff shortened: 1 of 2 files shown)


mattrose:
It should all be done in terminal.py in one method (spawn_child, IIRC) 

We should look at how other terminal emulators do the same thing, and copy them.  Also look at what launchers are expecting.  

I'm not too afraid of regressions at this point, as long as we properly document the command line options in the man page.

Issue #949, opened 2024-09-13
Fix the Save button behavior for Layouts


manueljordan:
When a custom layout is created or even when an already custom layout is edited, when the `Save` button is pressed it resets by complete all about:

* Profile
* Custom Command
* Working directory

Therefore all the settings are lost

So far the unique and correct approach to really save either the new or current layout is press the `Close` button

vssdeo:
Its reported in the issue that the save button in layout overrides the profile, custom command, working dir. The same does not happen when close button is pressed or pref window is closed. In my observation,I have seen that the profile is overwritten, but custom command and wording dir are updated.
    
The updation of working dir and custom command when editing without save button seems a bit different but I haven't changed that part for now.
    
For sake of consistency I have ensured the updation of profile (which anyways gets updated from drop down menu) but is over written by save button, happens when save button is pressed.

PR #970, opened 2024-12-11
-[bug 949]  #949 its reported in the issue that the save button in la…

vssdeo:
Its reported in the issue that the save button in layout overrides the profile, custom command, working dir. The same does not happen when close button is pressed or pref window is closed. In my observation,I have seen that the profile is overwritten, but custom command and wording dir are updated.

The updation of working dir and custom command when editing without save button seems a bit different but I haven't changed that part for now.

For sake of consistency I have ensured the updation of profile (which anyways gets updated from drop down menu) but is over written by save button, happens when save button is pressed.


mattrose:
Looks good, sorry about the wait.

Issue #971, opened 2024-12-11
Save button throws KeyError when a multiple tab layout is saved using Save Button after closing the Preference window using close button


vssdeo:
Launch terminator in debug mode -d

1. create a layout with 2 tabs.
2. save layout. Ensure you press close button in preferences and NOT (x) on top 
3. open preferences->layout again and selected the newly created layout with 2 tabs.
4. press save

you  will get the error below.

-whats happening is that on pressing close button in preferences.

            terminator.reconfigure() is called via on_closebutton_clicked(...)
            which in turn calls configure in notebook.py and clears
            self.last_active_term = {}

            on on pressing save point 4 above, that, eventually describe_layout in container.py gets called for Notebook and while setting it using self.get_nth_page, a KeyError is thrown since the terminal object was cleared as explained above.

            last_active_term.append(self.last_active_term[self.get_nth_page(tabnum)])

```
Traceback (most recent call last):
  File "/home/user/dev/terminator/terminatorlib/prefseditor.py", line 1615, in on_layoutrefreshbutton_clicked
    current_layout = terminator.describe_layout()
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/dev/terminator/terminatorlib/terminator.py", line 629, in describe_layout
    count = window.describe_layout(count, parent, layout, 0, save_cwd)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/dev/terminator/terminatorlib/container.py", line 336, in describe_layout
    count = child.describe_layout(count, name, global_layout, child_order, save_cwd)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/dev/terminator/terminatorlib/container.py", line 314, in describe_layout
    last_active_term.append(self.last_active_term[self.get_nth_page(tabnum)])
                            ~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^
KeyError: <terminal.Terminal object at 0x7fc933399a40 (terminatorlib+terminal+Terminal at 0x22195d0)>
```

vssdeo:
closing since KeyError may not affect the function it rather be exception handled and ignored.

PR #972, opened 2024-12-11
[bug 971] - Save button throws KeyError when a multiple tab layout is…

vssdeo:
Issue is described in #971 the KeyError should not be raised after following the steps.

The solution is to pass on flag so that the current terminals are not cleared on terminator.reconfigure when called from Preference window close button. Need to check if there is another efficient way to handle or parameter passing across to notebook.py is ok.

### DIFF
--- a/terminatorlib/notebook.py
+++ b/terminatorlib/notebook.py
@@ -53,7 +53,7 @@ class Notebook(Container, Gtk.Notebook):
 
         self.show_all()
 
-    def configure(self):
+    def configure(self, clear_last_active_term = True):
         """Apply widget-wide settings"""
         # FIXME: The old reordered handler updated Terminator.terminals with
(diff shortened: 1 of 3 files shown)


mattrose:
Q: Does the KeyError spit out affect the functionality at all, or is it just an error that is thrown because self.last_active_term[self.get_nth_page(tabnum)] is getting something that it's not expecting, but it doesn't actually affect the running of terminator as a whole. 

I'm just wondering if a better way of handling this error is to catch the KeyError when it's raised and just do the right thing, rather than alter the arguments to reconfigure.  This seems like the simpler way to fix this to me.

vssdeo:
@mattrose I think you are right, I guess I was thinking in terms of error handling and not considering that in python world exception need not be an error. I will remove this PR.

Issue #973, opened 2024-12-14
KeyError when a single window layout is selected after selecting a multiple tab layouts second terminal


vssdeo:
Launch terminator (2.1.4) in debug mode -d

    1. create a layout with 2 tabs.  and ensure the default layout has only one window.
    2. save layout via Add button in preferences
    3. select the 2 tab layout. (you can directly select a two tab layout if you already have using context layout menu)
    4. now first select the 2nd terminal of 2 tab layout AND now select the default layout
   
you will get the Key error below.

-whats happening is that on  selecting the single window layout, the system tries to save the single window layout but with previous settings of two tab layout. So we need to ensure that when the layout item is changed the layout sub menu also gets reset to first item.

```
ConfigBase::save: Processing layout: New Layout 2
ConfigBase::save: Processing layout: New Layout
noclass::get_config_dir: Found config dir: /home/user/.config
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/terminatorlib/prefseditor.py", line 1847, in on_layout_item_selection_changed
    self.layouteditor.on_layout_item_selection_changed(selection)
  File "/usr/lib/python3/dist-packages/terminatorlib/prefseditor.py", line 2147, in on_layout_item_selection_changed
    self.set_layout_item(item)
  File "/usr/lib/python3/dist-packages/terminatorlib/prefseditor.py", line 2152, in set_layout_item
    layout_item = layout[self.layout_item]
                  ~~~~~~^^^^^^^^^^^^^^^^^^
KeyError: 'terminal2'
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/terminatorlib/prefseditor.py", line 1847, in on_layout_item_selection_changed
    self.layouteditor.on_layout_item_selection_changed(selection)
  File "/usr/lib/python3/dist-packages/terminatorlib/prefseditor.py", line 2147, in on_layout_item_selection_changed
    self.set_layout_item(item)
  File "/usr/lib/python3/dist-packages/terminatorlib/prefseditor.py", line 2152, in set_layout_item
    layout_item = layout[self.layout_item]
                  ~~~~~~^^^^^^^^^^^^^^^^^^
KeyError: 'child1'
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/terminatorlib/prefseditor.py", line 1847, in on_layout_item_selection_changed
    self.layouteditor.on_layout_item_selection_changed(selection)
[...]

PR #974, opened 2024-12-14
Ensure that when the layout item is changed the layout sub item also gets reset to first item, else KeyError is thrown

vssdeo:
- KeyError when a single window layout is selected after selecting a multiple tab layouts second terminal as per issue #973 

- ensure that when the layout item is changed the layout sub item also gets reset to first item, else KeyError is thrown as described in bug


mattrose:
Looks good.  Sorry for the delay

PR #978, opened 2025-01-03
Searchbar improvements

pol-rivero:
- Fixed a bug where search is always initialized as case-sensitive, even if the "March Case" checkbox is unchecked.
- If search order is inverted, Shift+Enter will now search forward (instead of searching backward, which is the same direction as just pressing Enter).
- While we're at it, add some margin to the "Search:" label to improve how it looks

### DIFF
--- a/terminatorlib/searchbar.py
+++ b/terminatorlib/searchbar.py
@@ -43,6 +43,10 @@ class Searchbar(Gtk.HBox):
 
         self.config = Config()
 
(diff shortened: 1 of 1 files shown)


pol-rivero:
@mattrose could you take a look at this PR and #979?

mattrose:
Will do

pol-rivero:
Thanks! :)

mattrose:
Love it.  Thank you!

PR #979, opened 2025-01-03
Add Ctrl+Backspace keybind to delete word

pol-rivero:
Most users expect Ctrl+Backspace to delete the entire word placed to the left of the cursor. This PR adds a customizable keybind (defaults to Ctrl+Backspace) that does that.

### DIFF
--- a/doc/terminator_config.5
+++ b/doc/terminator_config.5
@@ -766,6 +766,13 @@ Reset the terminal state and clear the terminal window.
 Default value: \fB<Ctrl><Shift>G\fP
 .RE
 .sp
+\fBdelete_word\fP
+.RS 4
+Delete the word before the cursor.
+.br
+Default value: \fB<Ctrl>BackSpace\fP
+.RE
+.sp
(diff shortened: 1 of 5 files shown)


mattrose:
I'm wondering if this is not handled better by readline or the shell, not the terminal emulator.  Do you know if other terminal emulators do this?

mattrose:
I confirmed that this is handled by the readline library.   I really don't want to add functionality that is duplicated by another layer of the shell.  You should probably edit the readline configuration to get the behaviour you want in your shell.  I think all you're doing here is mapping ctrl+backspace to Meta-Delete and having that handled by readline.

Do you mind if I close this PR?

pol-rivero:
Ah I see, yes I'll close the PR.
Thank you for your time.

mattrose:
Thanks for understanding.  I hate turning away PRs, especially ones from people who have contributed successfully in the past, but I really want to keep the code as simple as I can.

Issue #984, opened 2025-01-15
Fix the check for VTE's notification-received patch


egmontkob:
See https://gitlab.gnome.org/GNOME/vte/-/issues/2858 for details. Quick summary:

The investigator of the crash said

> Ah, but terminator appears to be deliberately using this code as a feature test. [...] if the signal fails to connect, it falls back to assuming that that vte patch is unavailable.  It's just that it doesn't cope if vte actually crashes.

To which VTE's main developer responded

> Creating an instance and connecting to a signal to check if it exist isn't the right way to do this. The right way is to use `g_signal_lookup` with the name and gtype.

Although VTE was patched not to crash anymore, it would be nice if Terminator switched to the proper way of testing.

PR #986, opened 2025-01-15
Clean up VTE notification-received signal check

cjwatson:
This prompts a crash with some versions of VTE and Python (see https://gitlab.gnome.org/GNOME/vte/-/issues/2858), and although the crash has been fixed in VTE, VTE's main developer said that creating an instance and connecting to a signal isn't the right way to do this check.  Use the test they recommend instead.

Fixes: #984


mattrose:
Thanks for this!

PR #1002, opened 2025-04-02
Tmux-like navigation feature

siamak4mo:
In tmux, for example when you are at the leftmost pane and attempt to navigate to left, it simply navigates to the rightmost pane instead, so you can cycle through panes easily.
I have added this feature to the `navigate_terminal` function.

To test it, simply try to move in a direction (using key bindings) that reaches the edges of the terminal.


mattrose:
This is a big change to default behaviour, and I may end up pulling it, because it changes defaults, and kind of duplicates the  `cycle_next` and `cycle_prev` keybindings, but it's more intuitive, so I don't think it'll garner much complaint

PR #1028, opened 2025-10-24
fix: Use Menlo as default mono font on macOS

JAugusto42:
This PR fixes an issue where the default monospaced font renders poorly on macOS.

The previous logic relied on GNOME GSettings, which isn't available on macOS, causing it to use a messy fallback font. This change adds a check to detect if the OS is macOS (`Darwin`) and sets a better, platform-specific default font ("Menlo").

See the comparison below.
---
### Before:
Screenshot 2025-10-24 at 07 46 03

### After:
Screenshot 2025-10-24 at 07 49 17


mattrose:
Wow, I thought I was the only person crazy enough to use Terminator on a Mac.  I'll pull this in this evening.

PR #1039, opened 2025-12-19
Fix SaveLastSessionLayout by adding event-driven hooks

henricook:
I was looking around for another emulator because I thought my beloved Terminator lacked this, I'd never thought to check the plugins!

When trying this plugin out I found it worked for a single terminal, but didn't pay any attention to new tabs. These changes allow it to do that.

This is my first contribution, please let me know if you'd like any further work or demonstrative testing. I'm using this locally every day at the moment (starting today) so will keep an eye out for bugs.

### DIFF
--- a/terminatorlib/plugins/save_last_session_layout.py
+++ b/terminatorlib/plugins/save_last_session_layout.py
@@ -1,14 +1,16 @@
 import os
 import signal
-import sys 
+import sys
 
 # Fix imports when testing this file directly
 if __name__ == '__main__':
   sys.path.append( os.path.join(os.path.dirname(__file__), "../.."))
 
+from gi.repository import GLib
+
 from terminatorlib.config import Config
 import terminatorlib.plugin as plugin
-from terminatorlib.util import get_config_dir, err, dbg, gerr
+from terminatorlib.util import dbg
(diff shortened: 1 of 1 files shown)


terminatorlib/plugins/save_last_session_layout.py
```diff
@@ -1,82 +1,123 @@
...
-        dbg('signal handler called:signal %s (%s)' %
-                                        (signame, signum))
+        dbg('SaveLastSessionLayout: signal handler called: %s (%s)' % (signame, signum))
         self.save_session_layout()
-    
-    def connect_signals(self):
-        dbg("SaveLastSessionLayout connect_signals")
-
-        signal.signal(signal.SIGTERM, self.signal_handler)
-        signal.signal(signal.SIGCHLD, self.signal_handler)
```
henricook: This seemed like a noisy signal that we might not need as things inside open terminals are terminated

mattrose:
Looking at this now.  Looks good so far, I'm just going to run a few tests before merging

henricook:
Thanks Matt! I've been using it locally since I wrote it, every day, and not encountered any issues.

PR #1056, opened 2026-02-10
fix: convert enabled string to boolean in custom commands

### LINKED ISSUE #1047: Disabled CustomCommands keep appearing on right clik menu
**Describe the bug**
Disabled CustomCommands keep appearing. 
It looks like the problem is in here https://github.com/gnome-terminator/terminator/blob/master/terminatorlib/plugins/custom_commands.py#L76 , the "enabled" flag is left as string 'False' instead of being converted to boolean False.

**To Reproduce**
Create a command, disable it in the preferences, close all terminator instances, open a new one, the command will still be listed in the right click menu. When going to preferences the command will apear enabled again.

**Expected behavior**
The disabled commands should not appear on the right click menu.

 - OS: Ubuntu 24.04, 
 - Version_ 24.04
 - Display Technology: xfce

dmzoneill:
## Summary
- Fixes disabled custom commands appearing in the right-click menu
- Properly converts string 'False'/'True' values to boolean

## Details
The `enabled` flag from config was being stored as a string ('False' or 'True') instead of being converted to boolean. Since non-empty strings are truthy in Python, the expression `s["enabled"] and s["enabled"] or False` would evaluate to the string "False" for disabled commands, causing them to still appear in the menu.

The fix:
1. Uses `s.get("enabled", False)` for safer access
2. Explicitly checks if the value is a string
3. Converts string values ("true", "1", "yes") to boolean True, others to False

## Testing
- Verified the logic handles both string and boolean values correctly
- Disabled commands will now properly evaluate to False and not appear in the menu

Fixes #1047

### DIFF
--- a/terminatorlib/plugins/custom_commands.py
+++ b/terminatorlib/plugins/custom_commands.py
@@ -73,7 +73,9 @@ class CustomCommandsMenu(plugin.MenuItem):
         name = s["name"]
         name_parse = s.get("name_parse", "True")
         command = s["command"]
-        enabled = s["enabled"] and s["enabled"] or False
+        enabled = s.get("enabled", False)
+        if isinstance(enabled, str):
+            enabled = enabled.lower() in ("true", "1", "yes")
         if "position" in s:
           self.cmd_list[int(s["position"])] = {'enabled' : enabled,
                                                'name' : name,


mattrose:
Is this something you consider fixed?  Would you mind if I merged it?  I don't have much time to work on Terminator these days, but I do like getting and merging contributions from users when I do have time?  Re-opening.  Please have patience.  I do look at all PRs but don't have much time to properly inspect them.   For the next few months I have some time set aside to look at Terminator PRs, so hopefully I'll get around to this one soon

PR #1062, opened 2026-02-11
fix: convert enabled flag from string to boolean in custom commands

### LINKED ISSUE #1047: Disabled CustomCommands keep appearing on right clik menu

dmzoneill:
fixes #1047

the enabled config value is stored as a string but was being used directly as boolean, causing disabled commands to still appear in the right click menu. added proper string to boolean conversion to handle the 'False' string value correctly.

### DIFF
--- a/terminatorlib/plugins/custom_commands.py
+++ b/terminatorlib/plugins/custom_commands.py
@@ -73,7 +73,9 @@ class CustomCommandsMenu(plugin.MenuItem):
         name = s["name"]
         name_parse = s.get("name_parse", "True")
         command = s["command"]
-        enabled = s["enabled"] and s["enabled"] or False
+        enabled = s.get("enabled", "True")
+        if isinstance(enabled, str):
+            enabled = enabled.lower() in ('true', '1', 'yes')
         if "position" in s:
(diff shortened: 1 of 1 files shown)


mattrose:
Is this something you consider fixed?  Would you mind if I merged it?  I don't have much time to work on Terminator these days, but I do like getting and merging contributions from users when I do have time?  Re-opening.  Please have patience.  I do look at all PRs but don't have much time to properly inspect them.   For the next few months I have some time set aside to look at Terminator PRs, so hopefully I'll get around to this one soon

PR #1066, opened 2026-03-04
Fix: change tab title in tab where set_tab_tile is issued

### LINKED ISSUE #1018: set_tab_title for other than the active tab
I'm trying to use Claude Code hooks to write a status emoji in the tab title. I can find the current tab title, but set_tab_title seems to always use the currently selected tab. Are there a way to change the title for a tab by UUID or title?

A-725-K:
Fix #1018.

`set_tab_title` method now set the title to the tab which UUID matches the one passed to the method, instead of the current tab. Fallback on the current tab, in case the UUID does not match.

### DIFF
--- a/terminatorlib/ipc.py
+++ b/terminatorlib/ipc.py
@@ -309,7 +309,11 @@ class DBusService(Borg, dbus.service.Object):
             return
 
         notebook = window.get_children()[0]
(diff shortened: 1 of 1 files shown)


mattrose:
Thank you so much for this PR, I'm going away for the weekend but I hope to look at it next week.

PR #1067, opened 2026-03-04
Feat: Add keybinding for inserting terminal name

### LINKED ISSUE #960: Add a key binding for "Insert Terminal Name"
Since the Terminal Name is IMHO much more useful than the terminal number (which you cannot influence at all), it would be nice to have a shortsut for this as well. Since it is already in the menu I think the Implementation effort would be minimal.

A-725-K:
Fix #960.

Implement key binding for inserting the terminal name and set Super+2 as the default combination in the configuration. Add documentation.

### DIFF
--- a/doc/terminator_config.5
+++ b/doc/terminator_config.5
@@ -823,6 +823,13 @@ Rename the current terminal.
 Default value: \fB<Ctrl><Alt>X\fP
 .RE
 .sp
+\fBinsert_name\fP
+.RS 4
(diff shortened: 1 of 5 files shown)


mattrose:
Thank you so much for this PR, I'm going away for the weekend but I hope to look at it next week.

mattrose:
Looks good.  Thanks so much

mattrose:
Unfortunately to test this I need a non-mac keyboard.  I'll give it a shot soon

PR #1070, opened 2026-03-25
feat: tmux control mode with layout sync, capture, and pane management

ksanislo:
Add tmux -C (control mode) integration to Terminator, allowing it to act as a native GUI frontend for tmux sessions.

Core tmux integration (terminatorlib/tmux/):
- protocol.py: subprocess and PTY bridges for tmux control mode
- layout.py: parser for tmux layout strings into tree structures
- controller.py: per-session controller managing terminals ↔ panes
- handlers.py: layout-change, %output, window add/close/rename
- state.py: GObject state machine with signal-driven transitions

Layout synchronization:
- _apply_ratios maps tmux layout trees to GTK paned positions with cell-aligned boundaries (char_sep padding + round-up alignment)
- _snap_position uses math.ceil to match _apply_ratios grid
- _char_snap updated on VTE allocation for post-realization accuracy
- Expected allocations propagated to nested paneds to avoid stale GTK values during initial layout cascade
- Suppress spurious resize-pane during add-pane operations

Initial attach:
- capture-pane feeds existing content to VTE on attach
- Target-based settling ensures VTE sizes match tmux tree before capturing (active-window panes only)

Supporting changes:
- CLI: --tmux-attach, --tmux-new options
- IPC/DBus: tmux session routing for new windows
- Window: character-grid geometry hints, tmux-aware resize
- Config/prefs: overlay titlebar/scrollbar, auto-minimize origin
- Paned: cell-boundary snapping, tmux-managed handle tracking, do_size_allocate target restoration
- Docs: man pages and adoc for tmux options and config
- Tests: layout parser and protocol unit tests


ksanislo:
This is my attempt at expanding terminator with tmux integration support. It's still a bit too buggy to be considered yet, but it would be nice to get your thoughts on what you would want to see in here to allow this to be eligible for inclusion.

mattrose:
Wow, this is a lot.   I've able to read through about half the diff and so far it looks really good.  I'm going to need to sit with this a while and spend a good chunk of my rare free time with it before I can give you some feedback but I really like what you've done so far.

Issue #116, opened 2020-06-03
How to add a transparent blur to the background 


arshbot:
Transparency blurs are really cool, how can I get my terminator client to add a variable blur so I can't quite make out the background?

mattrose:
The closest thing I can think of is in the preferences window there's an option to shade the transparent background.  I don't know whether GTK or VTE supports a blurring effect.  If you have any ideas on how to do that I'd love to hear that.

br0kenbuild:
> Transparency blurs are really cool, how can I get my terminator client to add a variable blur so I can't quite make out the background?

being more specific is good - are you requesting individual terminal blur, independent off global one?
otherwise workaround would be 2 profiles with different global transparency levels and switching between them...

mayconritzmann:
This would really be an interesting feature, KDE Konsole implements this natively

mattrose:
theoretically it would be just adding another blur layer to the background.  PRs are welcome

martinsolberg:
I'd love to see this, like the Xfce Terminal Emulator does.

mattrose:
pycairo blur effect example code https://www.cairographics.org/cookbook/pythoncairopil/

CodingWithAnxiety:
Any update to this?

haasal:
I would like to implement this but don't really know where to start:

I found `draw_background()` in `terminal.py` but this seems only to be for the background image. Debug messages also aren't printed inside this function and breakpoints don't work which makes it really hard to debug. Help is appreciated :)

mattrose:
The background image is actually some good code to draw from.   The normal background is just a GTK RGBA object, so most of the work is done for us, including transparency.  The code for that is around here: https://github.com/gnome-terminator/terminator/blob/6d225c7aee3d42e37acac67629f7b321b7a4889b/terminatorlib/terminal.py#L747C1-L750C35
 
The `background_draw()` code draws the background image on top of the background RGBA object.  To blur, you would add another layer similar to draw_background that would just add the blur layer on top of the image, or the transparent background.  

On debugging, I assure you, debug messages work, hit me up on matrix and I'll talk you through it.

PR #1071, opened 2026-03-29
Background blur

Sotilrac:
## Summary

Adds a `background_blur` profile option that requests KDE KWin to blur the content behind transparent Terminator windows, creating a frosted-glass effect.

When `background_blur = True` and `background_type` is `transparent` or `image`, Terminator sets the `_KDE_NET_WM_BLUR_BEHIND_REGION` X11 window property via ctypes/libX11. KWin reads this property and applies its blur desktop effect behind the window.

Addresses Issue #116, but only for KDE.

### Changes

- **config.py**: Added `background_blur: False` to profile defaults
- **window.py**: Added `set_blur_behind()` method (ctypes X11 property) and `on_window_realize` signal to apply on startup
- **terminator.py**: Wired blur into the reconfigure cycle so toggling the preference takes effect immediately
- **prefseditor.py**: Added checkbox init, callback, and sensitivity logic
- **preferences.glade**: Added "Blur background (KDE)" checkbox in the Background tab

### Scope

- **KDE KWin on X11**: Full support via window property hint
- **Picom/other X11 compositors**: Property is safely ignored; users configure blur on the compositor side
- **Wayland**: Not implemented
- **GNOME/Mutter**: No per-window blur support in this compositor
- Default off, zero-cost when disabled, fails silently on unsupported setups

### Config example

```ini
[profiles]
  [[default]]
    background_type = transparent
    background_darkness = 0.8
    background_blur = True
```

### Testing

Tested on KDE Plasma / X11 with KWin blur desktop effect enabled. Verify with:

```bash
xprop -name Terminator | grep BLUR
# Should show: _KDE_NET_WM_BLUR_BEHIND_REGION(CARDINAL) = 0
```

### DIFF
--- a/terminatorlib/config.py
+++ b/terminatorlib/config.py
@@ -221,6 +221,7 @@ DEFAULTS = {
                 'background_image_mode' : 'stretch_and_fill',
                 'background_image_align_horiz': 'center',
                 'background_image_align_vert' : 'middle',
+                'background_blur'       : False,
                 'backspace_binding'     : 'ascii-del',
                 'delete_binding'        : 'escape-sequence',
                 'cursor_blink'          : True,
--- a/terminatorlib/preferences.glade
+++ b/terminatorlib/preferences.glade
@@ -3139,6 +3139,23 @@
                             <property name="position">6</property>
                           </packing>
                         </child>
+                        <child>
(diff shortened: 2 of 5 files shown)


mattrose:
I don't have a KDE env to test this on, but this all looks good.

PR #1080, opened 2026-04-25
Adds custom color selection and theme presets

vvbv:
Introduces new methods in the `Terminal` class to dynamically set background and foreground colors, including proper handling of transparency, inactive color offsets, and seamless updates to the VTE widget.

This enhancement is driven by the need to quickly differentiate between multiple terminal instances running similar or identical processes, without relying on predefined or static profiles for each one.

Additionally, the terminal's right-click context menu is extended with:

- A new "_Colors" submenu featuring a curated set of popular theme presets for quick selection.
- A "Custom..." option that opens a color picker dialog, allowing users to define precise background (with alpha support) and foreground colors on demand.

Overall, this provides a more flexible and immediate way to visually distinguish terminals, improving usability in workflows that involve multiple concurrent sessions.

### DIFF
--- a/terminatorlib/terminal.py
+++ b/terminatorlib/terminal.py
@@ -903,6 +903,60 @@ class Terminal(Gtk.VBox):
                 cursor_bg_color.parse(self.config['cursor_bg_color'])
             self.vte.set_color_cursor(cursor_bg_color)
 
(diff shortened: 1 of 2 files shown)


vvbv:
image

PR #1087, opened 2026-05-13
revert PR 1075

mattrose:
accidentally merged #1075 before it was ready


Duranzed:
What can be done for the plugin to be ready to use ?

PR #1091, opened 2026-06-02
Add session save and restore menu actions

vacu9708:
## Summary
Add terminal context menu actions for saving and restoring a lightweight session layout.
This introduces a new `Session` submenu in the terminal popup menu with:
- `Save session`
- `Restore session`

Saving persists the current Terminator layout using the existing layout serialization path, including each terminal’s current working directory. Restoring loads the saved session layout in the current process and replaces the window where restore was requested.

## Scope
This restores :
- windows/panes/tabs layout
- working directories

## Implementation Notes
- Saves the session as a reserved layout named `__last_session__`.
- Hides `__last_session__` from the normal Layouts submenu.
- Restoring creates the saved layout, then destroys the initiating window so the session replaces the current window rather than opening an extra one.

## Testing
- `python -m py_compile terminatorlib/terminator.py terminatorlib/terminal_popup_menu.py`
- `pytest -q terminatorlib/borg.py terminatorlib/config.py terminatorlib/cwd.py terminatorlib/factory.py terminatorlib/util.py tests/test_borg.py tests/test_signalman.py`

### DIFF
--- a/terminatorlib/terminal_popup_menu.py
+++ b/terminatorlib/terminal_popup_menu.py
@@ -7,7 +7,7 @@ from gi.repository import Gtk, Gdk
 
 from .version import APP_NAME
 from .translation import _
-from .terminator import Terminator
+from .terminator import LAST_SESSION_LAYOUT, Terminator
 from .util import err, dbg, spawn_new_terminator
 from .config import Config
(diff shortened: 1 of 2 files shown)


mattrose:
There's already a plugin that does this.

Issue #1008, opened 2025-05-28
Terminator prints '[x]' for binds Super+[x]


Areoform:
**Describe the bug**
Super+R prints 'r' instead of rotating terminals within the layout. This symptom is true of all Super+[x] keybinds within Terminator.

**To Reproduce**
1. Open a new Terminator process
2. Split vertically, then horizontally
3. Attempt rotation with Super+R, observe no rotation or other effects except 'r' printed to focused terminal

**Expected behavior**
Super+[x] binds activate their respective functions instead of printing character '[x]'

**Desktop (please complete the following information):**
 - OS or Linux Distribution: Arch Linux
 - Version: Rolling, up to date
 - Display Technology: Wayland
 - Terminator version(s): 2.1.4, 2.1.5-1
 - Dependency versions that may be relevant: vte3 (0.80.1-1), libkeybinder3 (0.3.2-5), xdg-utils (1.2.1-1), python-dbus (1.4.0-1)
 - DE: Plasma 6.3.5

**Additional context**
I can successfully use these functions if I keybind them to something else, e.g. Alt+R, and I can manually set the Meta+R keybind from within the application successfully, which displays it as Super+R as expected in both the UI and config file $home/.config/terminator/config

I've tested a handful of apps on my system, both native Wayland and running under XWayland, where I can successfully set and use Meta/Super+[key] keybinds (e.g. Kate, War Thunder)

I've confirmed there are no default DE keybinds creating conflicts, e.g. KDE Spectacle's default Meta+R or KDE Application Launcher's Meta - I even tried removing all DE keybinds possible through the Settings UI. To my knowledge the hidden Meta keybind to Application Launcher no longer exists in Plasma (confirmed with wev, special case handling of Meta goes away once unbound in Settings). Plasma 5.27 and 6.3 on other installs with default keybinds using Meta don't block these Terminator functions anyway.

I've check using wev that the correct keyboard events are generated by keypresses as expected.

I do not experience this problem with any other application or terminal emulator, neither the non-VTE Konsole nor a VTE based terminal emulator like Tilda.

Locale and keymap are set correctly. Terminator is running as a Wayland application as expected. Problem exists even when running Terminator without config file.

Confirmed working on:
    - Plasma 5.27.11 (Kubuntu 22.04 Jammy Jellyfish) with Terminator version 2.1.1-1 from repo jammy
[...]

PR #1095, opened 2026-06-16
Add mod4 key binding to keybindings.py

remijouannet:
quick addition to keybindings.py  for working around this issue https://github.com/gnome-terminator/terminator/issues/1008

it looks like super key in wayland has changed is keycode, it's now recognize has mod4 in gtk lib

this quick patch does not fix setting preference with the gui, but at least you can set your keybinding in the config file and it works

```
 next_tab = <Primary><mod4>semicolon
 prev_tab = <Primary><mod4>comma
```

### DIFF
--- a/terminatorlib/keybindings.py
+++ b/terminatorlib/keybindings.py
@@ -41,7 +41,8 @@ class Keybindings:
         'super':    Gdk.ModifierType.SUPER_MASK,
         'hyper':    Gdk.ModifierType.HYPER_MASK,
         'cmd':      Gdk.ModifierType.META_MASK,
-        'mod2':	    Gdk.ModifierType.MOD2_MASK
+        'mod2':	    Gdk.ModifierType.MOD2_MASK,
+        'mod4':     Gdk.ModifierType.MOD4_MASK,
     }
 
     empty = {}


mattrose:
Looks good.  I can't test this change as I only have apple keyboards at the moment.  If I could get more eyes on this it would be appreciated

mattrose:
@remijouannet what happens when you set the key in the preferences gui?  Does it set the keybinding to  <Primary><Mod4> even if you just press the super key?

remijouannet:
when i set the key in gui, it's recognized has <super>. but then the keybinding does not work, if i set mod4 in the configfile with my patch it works. 
i don't know if this is relevant but my keyboard is logitech mx keys (azerty), even if the issue seems to be related to wayland but i did not investigate.
i you want me to execute any command to diagnostic, let me know, my distro is fedora 44.

mattrose:
@remijouannet if you go into the preferences and set a previously unset keybinding, what is the resulting entry in the config file?

mattrose:
I'm wondering if it needs a fix similar to the PR #1093

remijouannet:
> @remijouannet if you go into the preferences and set a previously unset keybinding, what is the resulting entry in the config file?

well now, nothing happen, if i hit ctrl+super for example the keybinding stay unset.

however, interesting fact
`  next_tab = <Primary><mod4>semicolon`
is getting translate into this in gui

image

Issue #382, opened 2021-02-14
Display URLs behind links


nojhan:
With the use of OSC-8 escape sequences, an URL may be hidden behind any link text.
As such, it is impossible to have a clue on what the user is going to click, which is both breaking common UX assumptions and possibly adding a security problem.

A classical fix would be to display a transient tool-tip when the mouse is over the hyperlink, displaying the actual URL.
Another way would be to display the URL in the title bar.

nojhan:
A "copy URL" contextual menu command would also be nice to have.

mattrose:
I'll see if I can work up something.  The easiest thing might to have it show up in the context menu, but I'll take a stab at having it show up in a tooltip when you hover over it.  Copy URL should be fairly easy to add.

nojhan:
I would be willing to help on that issue, as I need it. What can I do to help? Where should I start?

mattrose:
The easiest way I can think of is to add the URL to the context menu when hovered over the link.  The code for generating the menu url stuff starts here https://github.com/gnome-terminator/terminator/blob/c447ae02016405ff69b20c198556c820baee14f2/terminatorlib/terminal_popup_menu.py#L49

PR #1099, opened 2026-06-29
Updates for context menu and urls.

ndunsworth:
Addresses #382 and adds open/copy menu items for OSC-8 url events as the current impl only adds them when vte.match_check_event succeeds.

When bulding the context menu check OSC-8 url events first via vte.hyperlink_check_event(event) then fall back to vte.match_check_event(event)

Set the open/copy menu item tooltips to the url string.  Preferred over including the url in the menu item name since a url could be massive.

### DIFF
--- a/terminatorlib/terminal_popup_menu.py
+++ b/terminatorlib/terminal_popup_menu.py
@@ -1,6 +1,6 @@
 # Terminator by Chris Jones <cmsj@tenshu.net>
 # GPL v2 only
-"""terminal_popup_menu.py - classes necessary to provide a terminal context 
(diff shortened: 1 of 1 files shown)


ndunsworth:
I need to fix an incorrect var name is why i pulled it

mattrose:
Ah, do you want me to close it again, or just mark it as a draft for now

ndunsworth:
That commit i just did should fix it, if not can switch to a draft.

I was bad and editing my installed python file while copying it into a clean clone of the repo :)

mattrose:
tysm for this!

Issue #1022, opened 2025-09-14
all instances crash when a tab is dragged outside of his window to create a new window with it


remi-remi:
**Describe the bug**
all instances crash when a tab is dragged outside of his window to create a new window with it
it doesn't happen all the time, i don't know why, but most of the time the crash occur.

**To Reproduce**
- launch terminator
- launch another instance (same arguments)
- create a new tab on instance 1
- drag it outside

**Expected behavior**
it shouldn't crash, especially the others terminator windows.

**Screen record** (with and without the -u)
https://github.com/user-attachments/assets/da8c8073-bea6-4cba-9708-158c3690b24d

**Desktop (please complete the following information):**

 - OS or Linux Distribution: [e.g. FreeBSD or Ubuntu]
Garuda Linux (Arch)
 - Version : rolling
 - Display Technology: Wayland

**Additional context**
$ terminator --version
terminator 2.1.5

$ python3 --version 
Python 3.13.7

$ cat /etc/os-release
NAME="Garuda Linux"
ID=garuda
ID_LIKE=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://garudalinux.org/"

$ uname -a
Linux darkArch 6.16.5-zen1-1-zen #1 ZEN SMP PREEMPT_DYNAMIC Thu, 04 Sep 2025 23:17:59 +0000 x86_64 GNU/Linux

# Session :
wayland KDE

$ pacman -Qi terminator vte3 vte-common gtk3 python python-gobject | egrep '^(Name|Version|Build Date|Packager)'
Name            : terminator
Version         : 2.1.5-1
Packager        : Christian Heusel <gromit@archlinux.org>
Build Date      : jeu. 22 mai 2025 09:47:03
Name            : vte3
Version         : 0.80.3-1
Packager        : Jan Alexander Steffens (heftig) <heftig@archlinux.org>
Build Date      : sam. 28 juin 2025 21:20:50
Name            : vte-common
Version         : 0.80.3-1
Packager        : Jan Alexander Steffens (heftig) <heftig@archlinux.org>
Build Date      : sam. 28 juin 2025 21:20:50
Name            : gtk3
Version         : 1:3.24.50-1
Packager        : Jan Alexander Steffens (heftig) <heftig@archlinux.org>
Build Date      : jeu. 07 août 2025 18:22:50
Name            : python
Version         : 3.13.7-1
[...]

PR #1101, opened 2026-07-06
Fix whole-app crash when dragging a tab out of a window; add explicit Detach Tab action

henricook:
# Background

Occasionally I drag a tab out of my terminator window on Gnome/Wayland and drop it, and it gets lost forever (but I think is still running!).

# What

- Fix dragging a tab out of a window crashing the whole app - every window at once (#1022)
- Add an explicit Detach Tab action: `Shift+Ctrl+D` keybinding + right-click menu item (tabs containing splits detach whole)
- Stop a failed tab drag leaving the snapped-back tab label undrawn on Wayland

# Why

The [gdb log attached to #1022](https://github.com/user-attachments/files/22393533/gdb.log) shows the segfault inside GTK's own drag-end handling:

```
#0  gtk_notebook_drag_end (...) at ../gtk/gtk/gtknotebook.c:3743
3743      _gtk_bin_set_child (GTK_BIN (priv->dnd_window), NULL);
```

GTK emits `create-window` in the middle of its drag bookkeeping, and our handler removed the page on the spot. With exactly two tabs, `hoover()` then dissolved the whole notebook, freeing `priv->dnd_window` before GTK's `drag-end` handler dereferences it (hence intermittent: 3+ tabs, no collapse, no crash). One Python process hosts every window, so "all instances crash".

The explicit action matters because on Wayland GTK only emits `create-window` for drops on shell chrome or bare desktop ([GTK#46](https://gitlab.gnome.org/GNOME/gtk/-/issues/46)) - dropping a tab on any window can never detach, and that's not fixable app-side.

# How

- `create_window_detach` now just schedules the detach with `GObject.idle_add`; the page is moved after the drag has fully ended, never inside the emission
- New `detach-tab` terminal signal wired like `move-tab`, driving the keybinding and menu item
- `queue_resize` after a failed tab drag so the restored label repaints ([GTK#3143](https://gitlab.gnome.org/GNOME/gtk/-/issues/3143))
- Verified on GNOME 50.1 Wayland via a nested compositor with injected drags: keyboard/menu/desktop-drop detach all work, drops on app windows snap back safely. Only test change is row indices in `test_prefseditor_keybindings.py` (the new binding sorts before `edit_*`)

Credit where due: Claude did the gdb spelunking and the nested-compositor testing.

### DIFF
--- a/terminatorlib/config.py
+++ b/terminatorlib/config.py
@@ -207,7 +207,8 @@ DEFAULTS = {
             'previous_profile' : '', 
             'preferences'      : '',
             'preferences_keybindings' : '<Control><Shift>k',
-            'help'             : 'F1'
+            'help'             : 'F1',
+            'detach_tab'       : '<Shift><Control>d'
         },
         'profiles': {
             'default':  {
--- a/terminatorlib/notebook.py
+++ b/terminatorlib/notebook.py
@@ -80,12 +80,7 @@ class Notebook(Container, Gtk.Notebook):
         self.last_active_term = {}
 
     def create_window_detach(self, notebook, widget, x, y):
-        """A tab has been dropped outside any tab bar. GTK emits this
-        signal in the middle of its drag-and-drop bookkeeping; removing
-        the page here (and potentially dissolving this notebook via
-        hoover) leaves GTK dereferencing freed drag state afterwards,
-        crashing the whole process (issue #1022). Defer the detach until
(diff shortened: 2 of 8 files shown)


mattrose:
I've been unable to reproduce, but I have seen this kind of behaviour in the past, and the code looks good to merge.  Can you report back here if the problem persists?

PR #1104, opened 2026-07-19
Consume Ctrl+click after opening a URL so it isn't opened twice

### LINKED ISSUE #1103: Ctrl+click opens URL twice when the foreground application also handles the click (mouse reporting)
### Summary

On a **Ctrl+click** on a URL, Terminator opens the link but does **not** consume the button-press event (`on_buttonpress` never returns `True`). GTK therefore keeps propagating the click, and VTE forwards it to the foreground application. Terminal applications that enable mouse reporting and treat Ctrl+click as "open link" (e.g. TUIs such as Claude Code, and comparable vim/tmux mouse setups) then open the **same URL a second time**, producing **two browser tabs** from a single click.

### Steps to reproduce

1. Run a TUI that (a) enables mouse reporting and (b) emits OSC‑8 hyperlinks and opens them on Ctrl+click. Any such app works; I reproduced with Claude Code.
2. Ctrl+click a link rendered by that app.
3. **Two** browser tabs open for the one click.

Control cases that confirm the mechanism:

- Ctrl+click an OSC‑8 hyperlink (or a plain URL) printed by a **normal shell** (no mouse reporting): exactly **one** tab. ✅
- The same link inside a mouse‑reporting TUI: **two** tabs. ❌

I instrumented this by pointing the default `http(s)` handler at a logging wrapper: a single click in the TUI records **two** launches of the same URL ~0.5–0.7 s apart — one from Terminator's `open_url`, one from the application.

### Root cause

`terminatorlib/terminal.py`, `on_buttonpress`, the `MOUSEBUTTON_LEFT` branch:

```python
if event.button == self.MOUSEBUTTON_LEFT:
    if self.config["link_single_click"] or event.get_state() & Gdk.ModifierType.CONTROL_MASK == Gdk.ModifierType.CONTROL_MASK:
        url = self.vte.hyperlink_check_event(event)
        if url:
            self.open_url(url, prepare=False)      # <-- no return True
        else:
            url = self.vte.match_check_event(event)
            if url[0]:
                self.open_url(url, prepare=True)    # <-- no return True
            else:
                dbg("No regex match, discard event.")
```

After Terminator opens a URL it should stop the event so the click it just consumed is not also delivered to the application underneath. Because the handler falls through returning a falsy value, the `button-press-event` signal keeps propagating to VTE's default handler, which (with mouse reporting on) forwards the click to the app.

### Proposed fix

Return `True` after a URL has actually been opened, so the event is consumed only in that case (normal clicks that don't hit a URL still fall through unchanged). PR to follow.

### Environment

- Terminator 2.1.5 (bug also present on current `master`)
- VTE-based, GTK3, Python 3.14
- Wayland session (XWayland), Manjaro Linux

aaronmhmr:
## What

On a **Ctrl+click** on a URL, `on_buttonpress` opens the link but never returns `True`, so GTK keeps propagating the `button-press-event`. With mouse reporting enabled, VTE forwards the same click to the foreground application. TUIs that treat Ctrl+click as "open link" (e.g. Claude Code, and comparable vim/tmux mouse setups) then open the **same URL a second time**, producing **two browser tabs** from a single click.

This returns `True` once a URL has actually been opened, so Terminator consumes the click it just handled. Clicks that don't hit a URL still fall through unchanged (selection, etc. are unaffected).

## Why it happens

`terminatorlib/terminal.py`, `on_buttonpress`, `MOUSEBUTTON_LEFT` branch — the two paths that call `open_url()` fall through without stopping the event, so the click reaches both Terminator's URL handler and the application underneath.

## Evidence

Pointing the default `http(s)` handler at a logging wrapper, a single Ctrl+click:

- in a **normal shell** (no mouse reporting): **1** browser launch ✅
- inside a **mouse-reporting TUI**: **2** launches ~0.5s apart — one from Terminator, one from the app ❌

With this change, the mouse-reporting case drops back to a single launch.

Fixes #1103

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/terminatorlib/terminal.py
+++ b/terminatorlib/terminal.py
@@ -1121,11 +1121,17 @@ class Terminal(Gtk.VBox):
                 dbg('url: %s' % url)
                 if url:
                     self.open_url(url, prepare=False)
+                    # Consume the event: the click we just acted on must not
+                    # also be forwarded to the application underneath. Apps that
+                    # enable mouse reporting and treat Ctrl+click as "open link"
+                    # (e.g. TUIs) would otherwise open the same URL a second time.
+                    return True
                 else:
                     dbg('OSC-8 URL not detected dropping back to regex match')
                     url = self.vte.match_check_event(event)
                     if url[0]:
                         self.open_url(url, prepare=True)
+                        return True
                     else:
                         dbg("No regex match, discard event.")
         elif event.button == self.MOUSEBUTTON_MIDDLE:


mattrose:
tysm for doing the work to fix this

PR #1111, opened 2026-08-17
Update immediately titlebar background color on D-Bus profile switch

dereckson:
When switching profile at runtime, for example through D-Bus using the command `remotinator switch_profile -p <profile>`, the titlebar color is not updated immediately. The color is only refreshed after the Terminator windows loses and regains focus.

Terminal.reconfigure() calls Titlebar.update() without providing the current focus context, so profile-dependent titlebar colors are not recalculated.

This change determines whether the toplevel window currently has input focus and update the titlebar accordingly.

### DIFF
--- a/terminatorlib/terminal.py
+++ b/terminatorlib/terminal.py
@@ -924,7 +924,15 @@ class Terminal(Gtk.VBox):
             elif self.config['scrollbar_position'] == 'right':
                 self.terminalbox.reorder_child(self.vte, 0)
 
-        self.titlebar.update()
+        toplevel = self.get_toplevel()
+        if isinstance(toplevel, Gtk.Window) and toplevel.has_toplevel_focus():
+            focused = self.terminator.get_focussed_terminal()
+            if focused is None:
+                focused = self.terminator.last_focused_term
+            self.titlebar.update(focused or self)
+        else:
+            self.titlebar.update('window-focus-out')
+
         self.vte.queue_draw()
 
     def set_cursor_color(self):


dereckson:
I've discovered this issue trying to switch from local to remote server profile, with two different backgrounds to give a hint where we are:

```
[...]

if [ -n "${TERMINATOR_UUID+1}" ]; then
        remotinator switch_profile -p "$SERVER"
        printf '\033]0;%s\007' "Servers   ::   $SERVER"
else
        printf '\033]0;%s\007' "$SERVER"
fi

ssh -t [...] tmux-reattach

if [ -n "${TERMINATOR_UUID+1}" ]; then
        remotinator switch_profile -p Local
        printf '\033]0;%s\007' "Local   ::   $HOSTNAME"
else
        printf '\033]0;%s\007' "$HOSTNAME"
fi
```
It's very convenient to update dynamically the profile. background and the title bar.

As I'm using a bordeau / lie-de-vin bar for local machine, and a teal bar for the remote server, I've noticed the issue and tried to fix it, hence this patch.

dereckson:
Can repro issue on:
  - Fedora RawHide on GNOME 51.beta, Wayland (my laptop)
  - Fedora 42 on Xorg on GNOME 48, X11 (my desktop)

On Wayland, the focus dance is more complicated to achieve: we need to do something forcing window redraw like set window to normal state and move it a little bit.

Can repro the fix on both environments too.

mattrose:
Fixes the problem nicely, tysm

